PR #35, opened 2021-09-08
Use `trybuild` for testing compilation failures

madsmtm:
Unsure if I want to add this, similar functionality is achievable using `compile_fail` doctests, though this is a lot better at testing proper error messages.


madsmtm:
I'm gonna go with this, because we're gonna need more complex macros (e.g. `derive` macros, see #55), and there it'll be important to test the output.

See also [this comment](https://github.com/madsmtm/objc2/pull/107#issuecomment-1013339232).

PR #46, opened 2021-10-15
Start using property testing

madsmtm:
Add `proptest` library to help with testing cases we haven't thought of.

- [x] Test `NSString`
- [x] Test `Class` getting
- [x] Test `Sel` registering
- [ ] ...


madsmtm:
Replaced by #74, fuzzing seems like the more robust approach overall

PR #48, opened 2021-10-28
`Id` conversion traits

madsmtm:
Proof-of-concept on fixing the remaining part of https://github.com/madsmtm/objc2/pull/41, identified in https://github.com/madsmtm/objc2/issues/39 (upstream https://github.com/SSheldon/rust-objc-foundation/issues/1).

Another alternative would be to mark `Id` with the unstable `#[fundamental]` attribute, see [RFC 1023](https://rust-lang.github.io/rfcs/1023-rebalancing-coherence.html) and [tracking issue](https://github.com/rust-lang/rust/issues/29635).

Issue #39, opened 2021-10-05
Consider moving `objc2_foundation` into the `objc2` crate


madsmtm:
There's a lot of places in `objc2_foundation` that would benefit from being able to implement `From` for specific `Id`s:
- `NSArray::from_vec`
- `NSArray::into_vec`
- `NSData::from_vec`
- `NSDictionary::from_keys_and_objects`
- `NSString::from_str`
- And so on...

Same goes for traits like `Default`.

But unfortunately, we can't create these implementations because `Id` is defined in another crate, see [RFC 2451](https://rust-lang.github.io/rfcs/2451-re-rebalancing-coherence.html). I tried creating a helper trait `FromId` for this purpose, but that doesn't work either:
```rust
pub trait FromId<T, O: Ownership> {
    fn from_id(obj: Id<T, O>) -> Self;
}

impl<T, O: Ownership, X: FromId<T, O>> From<Id<T, O>> for X {
    fn from(obj: Id<T, O>) -> Self {
        Self::from_id(obj)
    }
}

// Errors with:
// note: conflicting implementation in crate `core`:
// - impl<T> From<T> for T;
// note: downstream crates may implement trait `FromId<_, _>` for type `Id<_, _>`
```

Though doing it for `Default` would:

```rust
// In objc2:

pub trait IdDefault: Sized {
    type Ownership: Ownership;
    fn id_default() -> Id<Self, Self::Ownership>;
}

impl<T: IdDefault> Default for Id<T, T::Ownership> {
    fn default() -> Self {
        T::id_default()
    }
}

// In objc2_foundation:

impl IdDefault for NSString {
    type Ownership = <NSString as INSObject>::Ownership;
    fn id_default() -> Id<Self, Self::Ownership> {
        <NSString as INSObject>::new()
    }
}
```

So possible solutions:
- Move `objc2_foundation` to `objc2::foundation` (under a feature flag)
- Move `Id` to `objc2_foundation`?

PR #50, opened 2021-10-29
Make `Id` `#[fundamental]`

madsmtm:
Proof-of-concept on fixing the remaining part of https://github.com/madsmtm/objc2/pull/41, identified in https://github.com/madsmtm/objc2/issues/39 (upstream https://github.com/SSheldon/rust-objc-foundation/issues/1). Probably a better alternative than https://github.com/madsmtm/objc2/pull/48.

The `#[fundamental]` attribute is unstable, see [RFC 1023](https://rust-lang.github.io/rfcs/1023-rebalancing-coherence.html) and [tracking issue](https://github.com/rust-lang/rust/issues/29635), so I won't merge this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,7 +154,7 @@ jobs:
       # Use --no-fail-fast, except with dinghy
       TESTARGS: ${{ matrix.dinghy && ' ' || '--no-fail-fast' }} ${{ matrix.test-args }}
       FEATURES: ${{ matrix.features || 'malloc,block,exception,catch_all,verify_message' }}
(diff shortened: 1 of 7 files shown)


madsmtm:
So it turns out that since https://github.com/madsmtm/objc2/pull/150, we can't actually use this feature any more, it prevents us from implementing `MessageReceiver` for `Id` since downstream crates could `impl Message for Id<TheirType, O>` (which would be completely against the unsafe contract, but Rust can't know that!).

Hence I'll close this, if we find it to be a big issue we can always just use newtypes, like `NSArray<T>(Id<NSArrayInner<T>, Shared>)` if need be!

PR #69, opened 2021-11-03
Add `global_block!` macro

madsmtm:
Defines a `static` containing a `GlobalBlock`. This can be used directly in Objective-C (without the need for `_Block_copy`) and so can be quite convenient. (And is also more efficient).

TODO: ~Figure out whether the naming is proper (`fn` snake_case to `static` UPPER_CASE). See [C-CASE](https://rust-lang.github.io/api-guidelines/naming.html#c-case)~ Done.


madsmtm:
I went with the `static` syntax, because that better represents what the item ends up being

PR #73, opened 2021-11-03
Unstable: `extern type` support

madsmtm:
Requires `extern_types` feature from [RFC-1861](https://rust-lang.github.io/rfcs/1861-extern-types.html) and `ptr_metadata` feature.

I think I'm gonna wait with merging any of this until Rust progresses further in the design space.


madsmtm:
https://github.com/rust-lang/rust/issues/93959 helped a lot, https://github.com/rust-lang/rust/pull/101259 will probably put us where we can keep this in-tree

madsmtm:
I spent some more time a while back looking into making `ptr::cast` work with `T: PointeeSized + Thin`, it needs at the very least https://github.com/rust-lang/rust/pull/122553.

PR #81, opened 2021-11-23
Add `Id::retain_autoreleased`

madsmtm:
An optimized version of `Id::retain` that uses `objc_retainAutoreleasedReturnValue` for when the previous caller returns an autoreleased value with `objc_autoreleaseReturnValue`. This is a common thing in ARC code to avoid unnecessary `autorelease/retain` pairs, but it is (apart from changing the retain count) purely an optimization.

This uses the [newly stabilized](https://github.com/rust-lang/rust/pull/91728) inline assembly, which can easily cause _very_ undefined behaviour if used incorrectly. Luckily, the instructions are exceedingly simple (essentially just a special `nop`), so there's almost no room for error - so fingers crossed I got this right (at least enough that it doesn't cause UB)!

I've tested this on my local machine (both x86 and x86_64), and I've also added assembly tests to ensure that enough optimizations actually happen to make it work. At some point I would really like a helper like Apple's [`TestRoot`](https://github.com/apple-oss-distributions/objc4/blob/main/test/testroot.i) instead of the brittle checking of `retainCount` that we do now, but I'll defer that to later.

Related: https://github.com/gfx-rs/metal-rs/issues/222

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -306,6 +306,13 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: test
+        args: --no-default-features ${{ env.TESTARGS }} --release
+
+    - name: Test in release mode with features
+      if: ${{ !matrix.dinghy }}
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
         args: --features ${{ env.FEATURES }} ${{ env.TESTARGS }} --release
 
     - name: Run benchmarks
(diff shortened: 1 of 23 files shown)


madsmtm:
Since `objc_retainAutoreleasedReturnValue`, `objc_retain`, and so on return the same argument that they're given with, I briefly tested if it would be useful for LLVM to have this knowledge, encoded as follows.

```rust
if ptr != res {
    unsafe { core::hint::unreachable_unchecked() };
}
```

Conclusion: This was turned in to an [`@llvm.assume` intrinsic](https://llvm.org/docs/LangRef.html#llvm-assume-intrinsic), which can apparently have detrimental impact on performance.

madsmtm:
I considered just changing `Id::retain` to use `objc_retainAutoreleasedReturnValue`, but decided against it, even though most of the time you want to use `Id::retain_autoreleased`. Primary reasons was: to stay closer to Objective-C naming, and because #120 hopefully will mean most users don't have to think about it.

madsmtm:
Note: given that [Cranelift doesn't support inline assembly](https://github.com/bytecodealliance/wasmtime/issues/1041), [neither does `rustc_codegen_cranelift`](https://github.com/bjorn3/rustc_codegen_cranelift/issues/1204) yet, so that is broken.

A possibility would be to put the assembly stuff inside `Id::retain_autoreleased` behind a feature flag, I'm up for it if someone finds they need it.

Issue #85, opened 2021-12-03
`bindgen` support


madsmtm:
See https://github.com/rust-lang/rust-bindgen/issues/109.

I would like to create a (temporary?) fork of `bindgen` that generates bindings targetting `objc2`. In particular, `Encode` and `RefEncode` impls, but the `Message`/`MessageReceiver` stuff has also changed.

In the end, what I really want is the ability to generate _safe_ apis, but that's probably not gonna happen, it would at least require us to enrich every method with mutability information (`NSData -length` is immutable, while `NSMutableData -setLength:` is not), and we're never going to be able to generate a safe API for e.g. `NSData -initWithBytesNoCopy:length:deallocator:`.

madsmtm:
Maybe bindgen is really only useful for `-sys` crates. Though that shouldn't be undermined, it _would_ be a huge help instead of the current situation where you have to verify _every single argument_ by hand!

madsmtm:
Ideally the design of the bindings should make it easily possible to conform to our preferred design (that we haven't chosen yet, see https://github.com/madsmtm/objc2/issues/58).

Idea: In the end we have something like this:
```rust
pub struct NSBundle(ffi::NSBundle);

unsafe impl objc::Message for NSBundle {}

impl std::ops::Deref for NSBundle {
    type Target = NSObject;
    fn deref(&self) -> &Self::Target {
        NSObject::from_raw(*self.0) // Uses ffi::NSBundle's Deref impl
    }
}

impl NSBundle {
    pub fn new_with_path(path: &Path) -> Id<Self, Owned> {
        let nsstring = NSString::from(path);
        // NSString into ffi::NSString
        let obj = unsafe { ffi::NSBundle::alloc().initWithPath_(nsstring.into()) } as *mut NSBundle;
        Id::new(NonNull::new(obj).unwrap())
    }
    // new_with_url

    pub fn main() -> Option<&'static Self> { // Or Id or autorelease stuff
        let obj = unsafe { ffi::NSBundle::mainBundle() } as *mut NSBundle;
        unsafe { NonNull::new(obj).as_ref() }
    }

    pub fn path(&self) -> &NSString {
        unsafe { NSString::from_raw(self.0.bundlePath()) }
    }
}
```

Which uses the automatically generated bindings. The difference is as follows:

```diff
 #[repr(transparent)]
-pub struct NSBundle(NSObject);
+pub struct NSBundle(ffi::NSBundle);
 
[...]

PR #93, opened 2021-12-21
WIP on desired `bindgen` output

madsmtm:
Progress on evaluating if and/or how we should do #85.

Don't look too much at the `objc2-foundation-sys` stuff, the idea is that that part would be automatically generated - rather, focus on `objc2-foundation`, and think about whether using the bindings is better or worse.

### DIFF
--- a/objc2-foundation-sys/src/generated.rs
+++ b/objc2-foundation-sys/src/generated.rs
@@ -1,0 +1,236 @@
+//! Example of what I would like `bindgen` to be able to generate.
+
+use objc2::ffi::NSUInteger;
+use objc2::rc::{Allocated, Id, Unknown};
+use objc2::runtime::{Bool, Class, Imp, Object, Protocol, Sel};
+use objc2::{class, msg_send, msg_send_id, Encoding, Message, RefEncode};
+
+type NSInvocation = NSObject;
+type NSMethodSignature = NSObject;
+type NSString = NSObject;
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct _NSZone {
+    _unused: [u8; 0],
+}
+
+unsafe impl RefEncode for _NSZone {
+    const ENCODING_REF: Encoding<'static> = Encoding::Unknown;
+}
+
+#[repr(transparent)]
+pub struct NSObject(Object);
+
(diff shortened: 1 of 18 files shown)


madsmtm:
Another random idea: Add a new macro that makes it possible to still use the "familiar" `msg_send!` syntax:
```rust
msg_send_2![self, getObjects: vec.as_ptr(), range: range]
// ->
self.getObjects_range_(vec.as_ptr(), range)
```

This would require the [`paste`](https://github.com/dtolnay/paste) macro.

objc2-foundation-sys/src/generated.rs
```diff
@@ -0,0 +1,245 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl NSObject {
+    // alloc family
+
+    pub unsafe fn alloc() -> Option<Id<Self, Unknown>> {
+        Id::new_null(msg_send![class!(NSObject), alloc])
```
simlay, line 52: How are you thinking this might work with an inheritance structure? If the `alloc` message doesn't depend on the class, this is fine. My intuition tells me that it depends on the thing in `class!(...)` though.

madsmtm: I didn't originally see the reason for it, but what you've done is right (generate an `alloc` method on every class)

madsmtm: I realize I probably didn't formulate myself properly here... What I should have said is:

Yup, you're correct, all classes will need an `alloc` method. And that is also what I've done in this "proposal" / idea, `NSString::alloc`, `NSMutableString::alloc`, `NSData::alloc`, ... exists.

madsmtm: I added `ClassType::alloc` in #272, so that should be accounted for

madsmtm:
The biggest problem I had with this approach was that the separation between a `sys` crate and a "Rusty" crate lead to a lot of unnecessary duplication - and you're effectively always going to just be using the `sys` crate since we can't possibly hope to make "Rusty" bindings for all of Foundation/AppKit.

Future work in this problem space is happening in https://github.com/madsmtm/objc2/pull/264

Issue #25, opened 2021-09-05
Follow the Rust API Guidelines - Checklist


madsmtm:
# [Rust API Guidelines Checklist](https://rust-lang.github.io/api-guidelines/checklist.html)

- **Naming** *(crate aligns with Rust naming conventions)*
  - [x] Casing conforms to RFC 430 ([C-CASE])
  - [x] Ad-hoc conversions follow `as_`, `to_`, `into_` conventions ([C-CONV])
  - [x] Getter names follow Rust convention ([C-GETTER])
  - [x] Methods on collections that produce iterators follow `iter`, `iter_mut`, `into_iter` ([C-ITER])
  - [x] Iterator type names match the methods that produce them ([C-ITER-TY])
  - [x] Feature names are free of placeholder words ([C-FEATURE])
  - [x] Names use a consistent word order ([C-WORD-ORDER])
- **Interoperability** *(crate interacts nicely with other library functionality)*
  - [x] Types eagerly implement common traits ([C-COMMON-TRAITS])
    - `Copy`, `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`, `Hash`, `Debug`,
      `Display`, `Default`
  - [x] Conversions use the standard traits `From`, `AsRef`, `AsMut` ([C-CONV-TRAITS])
  - [x] Collections implement `FromIterator` and `Extend` ([C-COLLECT])
  - [x] Data structures implement Serde's `Serialize`, `Deserialize` ([C-SERDE])
    - I'm afraid of getting a circular dependency tree in the future if we do this (e.g. we rely on serde, and at some point, through their own deps, they rely on us)
  - [x] Types are `Send` and `Sync` where possible ([C-SEND-SYNC])
  - [x] Error types are meaningful and well-behaved ([C-GOOD-ERR])
  - [x] Binary number types provide `Hex`, `Octal`, `Binary` formatting ([C-NUM-FMT])
    - Doesn't make sense for `Bool`, just as we don't implement `PartialEq` for that
  - [x] Generic reader/writer functions take `R: Read` and `W: Write` by value ([C-RW-VALUE])
- **Macros** *(crate presents well-behaved macros)*
  - [x] Input syntax is evocative of the output ([C-EVOCATIVE])
  - [x] Macros compose well with attributes ([C-MACRO-ATTR])
  - [x] Item macros work anywhere that items are allowed ([C-ANYWHERE])
  - [x] Item macros support visibility specifiers ([C-MACRO-VIS])
[...]

PR #102, opened 2021-12-22
Better follow API guidelines

madsmtm:
Part of #25.

Contains a whole mixed bag of changes, see the changelogs and the diff.


madsmtm:
For the crate level docs vs. README, there's not really a concise guideline to follow, so I've chosen to keep most of it in the crate-level docs.

PR #104, opened 2022-01-02
Add static selectors

madsmtm:
My work on fixing upstream https://github.com/SSheldon/rust-objc/issues/49.

A lot of work has gone into this, by _many_ different people, see in particular the following:
- [`@pcwalton`'s investigation](https://github.com/SSheldon/rust-objc/issues/49#issuecomment-398526269)
- [`@mjbshaw`'s investigation](https://github.com/SSheldon/rust-objc/issues/49#issuecomment-416442267), their [Rust issue on statics and incremental compilation](https://github.com/rust-lang/rust/issues/53929) and their [comment on Rust's unstable `linkage` feature](https://github.com/rust-lang/rust/issues/29603#issuecomment-384420577)
- [`@jrmuizel`'s Rust issue](https://github.com/rust-lang/rust/issues/51608), similar to the above
- [`@clavin`'s discussion about their approach](https://github.com/SSheldon/rust-objc/issues/49#issuecomment-826056265)

This implementation is a combination of the following (`git` co-author attribution given where relevant):
- [`objrs`' implementation](https://gitlab.com/objrs/objrs/-/blob/b4f6598696b3fa622e6fddce7aff281770b0a8c2/macros/src/gen/sel_ref.rs)
- [`@mystor`'s implementation](https://github.com/mystor/rust-objc/tree/static_sel)
- [`@jrmuizel`'s implementation](https://github.com/jrmuizel/rust-objc/tree/static_sel) (fork of the above)
- [`@clavin`'s implementation](https://github.com/clavin/rust-objc/tree/static-refs)
- [`@nvzqz`'s implementation](https://github.com/nvzqz/fruity/commits/static-selectors) (haven't yet looked at this)

I've cut down heavily on procedural macro usage, because I think the code is much more readable without (ideally we'll find a way to get rid of them entirely).

TODO:
- [x] ~Set up ObjFW CI (https://github.com/madsmtm/objc2/pull/117)~ Postponed.
- [x] More test cases in CI, including at least different linking tests, testing in generic functions, and perhaps also UI tests (https://github.com/madsmtm/objc2/pull/35) and [LLVM/Assembly/...](https://github.com/Manishearth/compiletest-rs) output
- [x] ~Set up WinObjC CI~ Deferred
- [x] ~Make this work on GNUStep~ Deferred
- [x] Read LLVM, LLD and clang docs and sources
- [x] Read [objc4 sources](https://github.com/apple-oss-distributions/objc4) so see how the loading happens
[...]

PR #107, opened 2022-01-03
Use `compiletest_rs` for testing various properties of the compiler

madsmtm:
Similar to https://github.com/madsmtm/objc2/pull/35, except `compiletest_rs` is less polished than `trybuild`, so we have to do more work.

On the flip side, this it allows us to test codegen and assembly output, which we will probably need for https://github.com/madsmtm/objc2/pull/104 and others in the future.

See [The Guide to Rustc Development](https://rustc-dev-guide.rust-lang.org/tests/adding.html) for usage.


madsmtm:
Regarding codegen/assembly: That requires [`FileCheck`](https://llvm.org/docs/CommandGuide/FileCheck.html), which is a pain! In general, `compiletest_rs` is not really made for any of the things I'm trying to do here!

I'm strongly considering just going with #35 for the compile-fail tests, and rolling our own thing for testing codegen and assembly, by using the [`filecheck`](https://docs.rs/filecheck/0.5.0/filecheck/) and [`cargo_metadata`](https://crates.io/crates/cargo_metadata) crates instead (see also [`wasmtime`'s tests](https://github.com/bytecodealliance/wasmtime/blob/df37074218fe1e94a767e780e2138dde4dbd264c/cranelift/docs/testing.md#filecheck)).

madsmtm:
Replaced by #130

Issue #86, opened 2021-12-03
Better autorelease ergonomics


madsmtm:
The ergonomics of using autoreleased references are not great, especially when you are creating your own class. We should try to do better!

I see mainly three use-cases (for library / binding creators; end users probably won't touch this part of the library):
1. Returning a reference (shared in the example, could also be mutable)
    ```rust
    pub fn description<'pool>(&self, pool: &'pool AutoreleasePool) -> &'pool NSString {
        let desc: *const NSString = msg_send![self, description];
        pool.ptr_as_ref(desc)
    }
    ```
2. Exposing a retained version of the above:
    ```rust
    pub fn description_retained<'pool>(&self, pool: &'pool AutoreleasePool) -> Id<NSString, Shared> {
        // SAFETY: The NSString is valid and not mutably owned anywhere
        // The unsafe is required because the `&NSString` could have come from an `Id<NSString, Owned>`.
        unsafe { Id::retain(self.description(pool)) }
    }
    // Or
    pub fn description_retained(&self) -> Id<NSString, Shared> {
        autoreleasepool(|pool| {
            unsafe { Id::retain(self.description(pool)) }
        })
    }
    ```
3. Returning an autoreleased object in a custom / overridden method:
    ```rust
    extern "C" valid_attributes(_this: &Object, _sel: Sel) -> *mut NSArray {
        let array = Id<NSArray, Owned>::from(vec![...]);
        array.autorelease(pool) // Uhh, we don't have a pool from anywhere; it is implicit in Objective-C
    }
    let mut decl = ClassDecl::new("MyView", class!(NSView)).unwrap();
    decl.add_method(
        sel!(validAttributesForMarkedText),
        valid_attributes as extern "C" fn(&Object, Sel) -> *mut NSArray,
    );
    ```

See related: https://github.com/gfx-rs/metal-rs/issues/222

madsmtm:
Ideas for making use-case 2 possible without `unsafe`:
- Add `Autoreleased<'pool, T, O: Ownership>` helper struct. The ergonomics is unfortunately a bit degraded, since users will have to learn yet another struct. Also not sure what we should do regarding `Encode`/`RefEncode` on this?
- Alternative: `Autoreleased<T>`, used as `&'pool [mut] Autoreleased<T>`? Not sure if it would help with the safety.

Ideas for use-case 3:
[...]

Issue #30, opened 2021-09-05
Better class declaration


madsmtm:
~Blocked on at least https://github.com/madsmtm/objc2/pull/21.~

Fundamentally cannot be made safe, since you're calling into unknown Objective-C classes. Not even sure `add_ivar` is sound (since Objective-C could be using the same ivar in a superclass)?

- [x] Fix https://github.com/SSheldon/rust-objc/issues/12
  Fixed by https://github.com/madsmtm/objc2/pull/193.
- [x] Rename `ClassDecl` -> `ClassBuilder`
- [x] Can it be made more ergonomic with macros? See https://github.com/SSheldon/rust-objc/issues/74.
  Done in various PRs (primarily `declare_class!` macro)
- [x] Returning an autoreleased object in a custom / overridden method is very unergonomic:
    ```rust
    extern "C" valid_attributes(_this: &Object, _sel: Sel) -> *mut NSArray {
        let array = Id<NSArray, Owned>::from(vec![...]);
        array.autorelease(pool) // Uhh, we don't have a pool from anywhere; it is implicit in Objective-C
    }
    let mut decl = ClassDecl::new("MyView", class!(NSView)).unwrap();
    decl.add_method(
        sel!(validAttributesForMarkedText),
        valid_attributes as extern "C" fn(&Object, Sel) -> *mut NSArray,
    );
    ```

    Will also need something like an `autorelease_return` method on `Id` which calls `objc_autoreleaseReturnValue`.

    Fixed by https://github.com/madsmtm/objc2/pull/191. See https://github.com/madsmtm/objc2/pull/112 as well for an alternate solution.
- [x] ~Differentiate between static class references and "registered" classes, see https://github.com/SSheldon/rust-objc/issues/49#issuecomment-830429031~ Moved to https://github.com/madsmtm/objc2/pull/185.
- [x] Memory management rules are not automatically followed. Fixed by https://github.com/madsmtm/objc2/pull/244
- [x] Uninitialized data in `init` methods. Fixed by https://github.com/madsmtm/objc2/pull/252
- [x] Allow `Drop` types in ivars. Fixed by https://github.com/madsmtm/objc2/pull/254

madsmtm:
Work on syntax extension that would help with declaring new classes (ideas in part from `objrs`, example from GNUStep's `GSTitleView`):

```rust
unsafe pub struct MyTitleView: NSView {
    pub close_button: Option<Id<NSButton, Owned>>,
[...]

PR #112, opened 2022-01-06
Add `autoreleasepool_leaking`

madsmtm:
Idea for https://github.com/madsmtm/objc2/issues/86 and/or https://github.com/madsmtm/objc2/issues/30.

Another idea for this:

```rust
impl AutoreleasePool {
    const LEAKING: &'static Self = {
        &*ManuallyDrop::new(unsafe { Self::new_leaking() })
    };
}
```

But that would always yield a static lifetime, which would mean it had to be `unsafe` somehow!

### DIFF
--- a/objc2/src/rc/autorelease.rs
+++ b/objc2/src/rc/autorelease.rs
@@ -1,7 +1,8 @@
-use core::cell::UnsafeCell;
 use core::ffi::c_void;
 use core::fmt;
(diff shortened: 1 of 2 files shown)


madsmtm:
I'm going to close this since #191 added `Id::autorelease_return` which basically covers all the use-cases this would've had.

PR #113, opened 2022-01-06
Add a `terminated` dependency and use `NulTerminatedStr` for inputs

madsmtm:
`@SSheldon`'s original work on SSheldon/rust-objc#54.


madsmtm:
I don't think this makes sense to do, since actually, none of these APIs enforce valid UTF-8 strings.

Rather, we should just expose the inherent `CStr`-nature of these APIs some more (which becomes feasible from an ergonomics standpoint after https://github.com/rust-lang/rust/issues/105723)

PR #114, opened 2022-01-06
Move `Encode` types to exported module

madsmtm:
In similar vein to the `ffi` module.

This is a rather large breaking change, we'd probably want to hold off on doing this


madsmtm:
Replaced by https://github.com/madsmtm/objc2/pull/135

Issue #58, opened 2021-11-02
Downgrading using `Deref`


madsmtm:
We often want to send messages to methods defined on the superclass; in Objective-C this is automatically supported by the runtime, and we utilize this in `objc2_foundation` by creating traits for each superclass `INSObject`, `INSString`, ... and implement them for our object.

## The problems

- Usage is more cumbersome for the user, since they have to import both the object they want to use _and_ all the traits for the superclasses. E.g. when using `NSMutableArray` you often have to do:
  ```rust
  use objc2_foundation::{NSMutableArray, INSMutableArray, INSArray, INSObject};
  ```
  (Could be solved by adding a `prelude` module which exports these traits).

- When defining functions taking an object the creator has to remember to create a generic function over the helper trait, instead of just taking the object they were interested in:
  ```rust
  use objc2_foundation::{INSObject, INSString, NSString};
  
  fn my_fn(obj: &impl INSObject) { // Instead of just `obj: &NSObject`
      println!("{}", obj.description())
  }
  
  let obj = NSString::new();
  my_fn(&obj);
  ```

- The traits contain default functions (e.g. `INSObject::description`), which are then made generic on use (the compiler creates `NSObject::description`, `NSString::description`, `NSArray::description`, ...). These functions are identical however, leading to an increase in code-size.

## The solution

Rust does not have inheritance, but it has something that can emulate it; `Deref`/`DerefMut`! Using these, we can "downgrade" objects, e.g. convert `&NSMutableString` to `&NSString`, and that to `&NSObject`, and use methods defined on `NSObject`.

This is not a new idea, `fruity` uses this approach, see its [`subclass!`](https://github.com/nvzqz/fruity/blob/main/src/core/macros.rs#L4-L63) and [`objc_subclass!`](https://github.com/nvzqz/fruity/blob/main/src/objc/macros.rs#L34-L61) macros, though they don't create mutable references to objects. `objrs` [does it as well](https://gitlab.com/objrs/objrs/-/blob/b4f6598696b3fa622e6fddce7aff281770b0a8c2/macros/src/gen/gen_class.rs#L136-188) in their `#[objrs(class, ...)]` macro.

## Safety

Since objects are always just pointers into heap, and `msg_send!` doesn't care if the given pointer has one type or the other, the pointer casting part should be safe. However, we also have to ensure Rust's safety invariants are upheld, especially in relation to unique/aliased references and lifetimes.

[...]

PR #115, opened 2022-01-06
Use `Deref` in `objc2-foundation` and remove "`INS`" helper traits

madsmtm:
Huge part of https://github.com/madsmtm/objc2/issues/58.

Also renamed protocols, now that the possible confusion from them is gone.

(Yes, there exists both the `NSObject` class and the `NSObject` protocol, but that is something we'll paper over).


madsmtm:
I'm probably just gonna go and merge this, though we need a better solution for specifying a bound that says "this is an object, and not just something that is messageable" (see also https://github.com/madsmtm/objc2/pull/66).

PR #117, opened 2022-01-06
`objfw` support

madsmtm:
WIP


madsmtm:
ObjFW [recently reached 1.0](https://objfw.nil.im/wiki?name=Releases), so I'll probably add support for that soon

PR #119, opened 2022-01-11
Set up iOS testing

madsmtm:
Using [`dinghy`](https://github.com/sonos/dinghy).

Getting this to run on my old iPad mini 1st generation was a real pain, especially because my macOS version is slowly getting more and more unsupported, so XCode was having trouble with certificates. But hey, got it to work!

See also:
- [Compile Rust tests with cargo for the iOS simulator](https://gist.github.com/learnopengles/60c945b92fbe60a0fecaa0143c35c4af)
- [Testing Rust on iOS with Travis](https://sasheldon.com/blog/2016/02/22/testing-rust-on-ios-with-travis/)


.github/workflows/ci.yml
```diff
@@ -233,13 +247,32 @@ jobs:
...
+        # Launch the simulator
+        xcrun simctl list runtimes
+        RUNTIME_ID=$(xcrun simctl list runtimes | grep iOS | cut -d ' ' -f 7 | tail -1)
+        export SIM_ID=$(xcrun simctl create My-iphone7 com.apple.CoreSimulator.SimDeviceType.iPhone-7 $RUNTIME_ID)
+        xcrun simctl boot $SIM_ID
+
+        # Overwrite `cargo` to point to `cargo-dinghy`
+        mkdir bin
+        echo "#!/bin/bash" > bin/cargo
+        echo "cargo-dinghy --device=$SIM_ID \"\$@\"" >> bin/cargo
```
simlay, line 264: I see why you did this but think it may save you future time to just have the `cargo dinghy test` section in this run-block and below in the `Test without features` just skip it it with an `if: ${{ !matrix.dinghy }}`. Overriding the  `cargo` with a new shell-script cargo feels a bit fragile (maybe the `actions-rs/cargo` action changes to not look at `PATH` for it's cargo) and isn't something a lot of people would expect.

helper-scripts/test-local.fish
```diff
@@ -0,0 +1,38 @@
...
+cargo dinghy --device sim test
+cargo dinghy --device sim test --release
+
+# Test on my iPad mini 1st generation iOS 9.3
+# Followed this guide to set it up:
+# https://github.com/sonos/dinghy/blob/main/docs/ios.md
+#
+# We use build-std because the target is armv7-apple-ios
+mkdir .cargo
+echo '[unstable]' > .cargo/config.toml
```
simlay, line 26: I'm mildly surprised that the [`cargo` manifest doesn't have target specific unstable features](https://doc.rust-lang.org/cargo/reference/unstable.html).

simlay (commented):
If [futureware-tech/simulator-action@v1](https://github.com/marketplace/actions/launch-ios-simulator`) had more activity, I'd suggest using it but at this point, a `run` block seems best for starting an iOS simulator.

helper-scripts/test-local.fish
madsmtm: Yeah, that would have been useful!

Just for reference, the reason I did it this way here is because `cargo dinghy` doesn't support `-Zbuild-std`.

.github/workflows/ci.yml
madsmtm: Yeah you're right, I did it like this because I was hoping the other scripts would work as well, but it turns out `cargo dinghy` doesn't support specifying features from a workspace root, so it's kind of useless like this anyhow.

simlay: Oh, that makes more sense. I was thinking you did it for ease of use with the `actions-rs/cargo` action.

> it turns out cargo dinghy doesn't support specifying features from a workspace root, so it's kind of useless like this anyhow.

Yeah. Dinghy isn't perfect. When I added [fixed iOS simulator support](https://github.com/sonos/dinghy/pull/96), I wasn't in love with how I did it but it's better than nothing. Looking back, I'm glad I added it as I see more and more iOS CI.

Issue #87, opened 2021-12-10
Add `rc::Unknown` and `rc::MaybeOwnership`


madsmtm:
Rough sketch:

When generating bindings to Objective-C classes automatically (e.g. with `bindgen`, see #85), it would be nice to return a reference counting pointer (e.g. an `Id`). Sadly, however, it is not possible to know the ownership of that without inspection of the entire class; even handing out immutable references can easily trigger undefined behaviour (e.g. if there was a `Id<T, Owned>` somewhere else in the program).

Therefore it would be beneficial to add a reference-counting pointer that does not guarantee any form of ownership, like the `StrongPtr` removed in https://github.com/madsmtm/objc2/commit/e08e87d210dc8d1a7f77a0bffe7921076b0fc40d. See also upstream https://github.com/SSheldon/rust-objc/issues/24.

This can probably be accomplished by adding a `rc::Unknown`, similar to `rc::Owned` and `rc::Shared`.
On initial inspection, you would just implement `rc::Ownership` for `rc::Unknown`, but that makes all high-level code unable to deal with the "normal" case where you want to know the ownership, so we should consider adding `rc::MaybeOwnership` as well to deal with this case.

Example code changes:
```rust
pub trait MaybeOwnership: private::Sealed + 'static {}
pub trait Ownership: MaybeOwnership {}

impl MaybeOwnership for Unknown {}
impl MaybeOwnership for Owned {}
impl MaybeOwnership for Shared {}
impl Ownership for Owned {}
impl Ownership for Shared {}


pub struct Id<T, O: MaybeOwnership> { ... }

impl<T: Message> Id<T, Unknown> {
    pub unsafe fn into_shared(self) -> Id<T, Shared> { ... }
    pub unsafe fn into_owned(self) -> Id<T, Owned> { ... }
}

// Still only `Owned` and `Shared` are `Deref`!
impl<T, O: Ownership> Deref for Id<T, O> { ... }

// `Unknown` can also be sent messages (since Objective-C doesn't care about ownership)
unsafe impl<T: Message, O: MaybeOwnership> MessageReceiver for Id<T, O> { ... }
```

madsmtm:
Not that it matters much, but this could be accomplished without breaking changes.

madsmtm:
`Id<MaybeUninit<T>, Unknown>` is not really possible, since the type is actually partially initialized, just not entirely. Something like `Id<Alloc<T>, Unknown>` would be better for that.

[...]

PR #120, opened 2022-01-12
Add `msg_send_id!`

madsmtm:
Add `msg_send_id!` to help with following [memory management rules](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmRules.html#//apple_ref/doc/uid/20000994-BAJHFBGH) (see also [this](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#retain-count-semantics)).

Closes https://github.com/madsmtm/objc2/issues/86.

TODO:
- [x] Figure out https://github.com/madsmtm/objc2/pull/81.
- [x] Figure out https://github.com/madsmtm/objc2/pull/150.
- [x] Better testing, ~blocked on https://github.com/madsmtm/objc2/pull/166~.
- [x] Proper initial implementation
- [x] Should `alloc` return `Option<Id<Allocated<T>, O>>`, and just allow passing that to `init`, or should we handle it?
  It should and now does.
- [x] Test whether diagnostics are better if we use associated types (means that `__MsgSendId`/`__MsgSuperSendId` can only be implemented for one type pr. `__Allocated<...>`, which may be better?)
  - I've tweaked things and tried a few different options, what we have now is pretty good!
- [x] ~Consider using `MaybeUninit` instead of `Allocated`?~
  Nope, see https://github.com/madsmtm/objc2/issues/87#issuecomment-1148099411
- [x] ~`super` calls~ Postponed, to https://github.com/madsmtm/objc2/pull/173
- [x] Moved `rc::Allocated<T>` to https://github.com/madsmtm/objc2/pull/172 to get this merged.
- [x] Consider whether `init` should take generic `Into<Option>`, or just `Option`.
- [x] Documentation
- [x] Cleanup

### DIFF
--- a/objc2-foundation/examples/class_with_lifetime.rs
+++ b/objc2-foundation/examples/class_with_lifetime.rs
@@ -5,7 +5,7 @@ use std::sync::Once;
 use objc2::declare::ClassBuilder;
 use objc2::rc::{Id, Owned, Shared};
 use objc2::runtime::{Class, Object, Sel};
-use objc2::{msg_send, sel};
+use objc2::{msg_send, msg_send_id, sel};
 use objc2::{Encoding, Message, RefEncode};
 use objc2_foundation::NSObject;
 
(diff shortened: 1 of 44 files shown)


madsmtm:
Another, less versatile idea, add this (and similar) to `msg_send!`:
```rust
($obj:expr, new) => ({
    let sel = $crate::sel!(new);
    let result: Option<$crate::rc::Id<_, _>>;
    match $crate::MessageReceiver::send_message(&$obj, sel, ()) {
        Err(s) => panic!("{}", s),
        Ok(Some(r)) => result = Some($crate::rc::Id::new(r)),
        Ok(None) => result = None,
    }
    result
});

// Id::new(msg_send![class!(NSObject), new])
// Now becomes
// msg_send![class!(NSObject), new].unwrap()
```

Issue #125, opened 2022-01-19
Make `msg_send` automatically convert `BOOL` to `bool`


ArturKovacs:
As far as I understand, one always has to specify `BOOL` instead of `bool` as the return type of messages. This is error prone.

But objc doesn't have the `bool` type, so it seems like a safe assumption that whenever the return (or parameter) type is specified as `bool` it should instead be `BOOL`. With this assumption, `msg_send` could do the conversion internally. 

Right now we have to write
```rust
let my_var: BOOL = msg_send![my_object, my_method];
if my_var.as_bool() {
    // stuff
}
```

But it would be much safer if we could write the following, and `msg_send` would do the correct conversion automatically.
```rust
let my_var = msg_send![my_object, my_method];
if my_var {
    // stuff
}
```

madsmtm:
I like the idea, though I don't really think it's possible?

`msg_send!` effectively desugars to a generic function call (generic over both return type and arguments):
```rust
let my_var = msg_send![my_object, my_method];
if my_var {}

// Effectively becomes

let my_var = MessageReceiver::send_message(my_object, sel!(my_method), ());
if my_var {}
```

Somehow we would have to put a conversion step in there, an idea could be like this:
```rust
pub trait IntoBoolIfBool {
    type BoolType;
    fn into_bool_if_bool(self) -> Self::BoolType;
}

impl IntoBoolIfBool for runtime::Bool {
    type BoolType = bool;
    fn into_bool_if_bool(self) -> Self::BoolType {
        self.as_bool()
    }
}

impl IntoBoolIfBool for i32 {
    type BoolType = i32;
    fn into_bool_if_bool(self) -> Self::BoolType {
        self
    }
}

let my_var = MessageReceiver::send_message(my_object, sel!(my_method), ());
let my_var_converted = IntoBoolIfBool::into_bool_if_bool(my_var);
if my_var_converted {}
```

But since we can't tell the type system that there is only one implementor of `IntoBoolIfBool` that returns `bool`, the type of `my_var` would probably still have to be annotated?

Other solutions (that I think would be possible):
1. Disallow returning `bool` from `msg_send!`.
  
    If we still want to allow passing `bool` as a parameter, this would require a new `trait EncodeReturn: Encode` (related to https://github.com/madsmtm/objc2/issues/118) for types that can safely be constructed from Objective-C (note: I'm reluctant to add too many new traits, since it means downstream library creators have more traits to implement; but this one _may_ be worth it).
[...]

PR #126, opened 2022-01-19
WIP: Allow `bool` as the return type from `msg_send!`

madsmtm:
A proof of concept for https://github.com/madsmtm/objc2/issues/125.

There's an issue with `objc2-encode` being its own crate, which means we can't write the `impl<T: Encode> ReturnType for T` in any other crate, but that's solvable (by merging `objc2` and `objc2-encode`, though not very nice).

Instead of the `impl<T: Encode> ReturnType for T` impl, what we really want is specialization :/

### DIFF
--- a/objc2-encode/src/encode.rs
+++ b/objc2-encode/src/encode.rs
@@ -183,13 +183,13 @@ unsafe impl Encode for () {
     const ENCODING: Encoding<'static> = Encoding::Void;
 }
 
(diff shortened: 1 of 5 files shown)


ArturKovacs:
This is awesome!

madsmtm:
FYI: Picked this up again in https://github.com/madsmtm/objc2/pull/239, extended to parameter types as well as return types

PR #127, opened 2022-01-19
Verify that an `Id`s' ownership is correct using associated objects

madsmtm:
Resolves upstream https://github.com/SSheldon/rust-objc-id/issues/5.

A downside to this is that `ManuallyDrop::new(id)` no longer works properly, you'd have to use `id.forget()`, which is less ergonomic.

I do think this is a good idea, it helped me catch a bug in `NSArray::new` (`NSArray::new` always returns the same object), fix included in #163.

TODO:
- [ ] Documentation
- [ ] ...


madsmtm:
I think I may put this behind an unstable feature that you'd enable while testing, to move it forwards for now!

madsmtm:
Ownership is now tracked in a substantially different way, and usually things won't be mutable, so I don't think this has enough value to be worth the cost.

PR #128, opened 2022-02-22
Soundness fixes

madsmtm:
- Properly mark objects as mutable behind shared references.
- Fix auto trait impls, see https://github.com/rust-lang/rust/issues/93367.

### DIFF
--- a/block-sys/src/lib.rs
+++ b/block-sys/src/lib.rs
@@ -35,7 +35,7 @@ pub struct Class {
     _priv: [u8; 0],
 
     /// See objc_sys::OpaqueData
-    _opaque: PhantomData<(UnsafeCell<()>, *const UnsafeCell<()>, PhantomPinned)>,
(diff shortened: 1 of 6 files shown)


madsmtm:
Didn't really realize this before, but making `type OpaqueData = UnsafeCell<...>` actually means that `Id<T, Shared>` is much safer!

Having aliasing `Id<T, Owned>` and `Id<T, Shared>` is still UB, but if you only use `Id<T, Shared>` and don't implement `Send` and/or `Sync` on your type then Objective-C methods are free to mutate the object as much as they want!

madsmtm:
To consider: Does it make sense for `Id` to store an `UnsafeCell`?

madsmtm:
Also, the pointer in `Id<Object, Shared>` may actually have been mutated by `Id::clone`, so this is definitely better!

PR #136, opened 2022-03-02
Try using the `allocator_api` to create objects

madsmtm:
Just a fun idea for making `Box<Object> ~= Id<Object, Owned>`.

One _could_ probably make it work better than I have here, but there are some pretty fundamental issues that are not easily fixed, like `Layout` being entirely unfit for this application. Maybe [making `Layout` an associated type of `Allocator`](https://rust-lang.github.io/rfcs/1398-kinds-of-allocators.html#make-layout-an-associated-type-of-allocator-trait) could do the trick, maybe not.

### DIFF
--- a/objc2/src/lib.rs
+++ b/objc2/src/lib.rs
@@ -125,6 +125,7 @@
     feature = "unstable_autoreleasesafe",
     feature(negative_impls, auto_traits)
 )]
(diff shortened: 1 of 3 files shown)


madsmtm:
Anyhow, definitely not gonna merge this, was just to see if it could be done

PR #142, opened 2022-04-04
Runtime selection using crate features in "higher level" crates

### LINKED ISSUE #140: Choose sensible default runtime in `-sys` libraries?
Currently:
- `block-sys` chooses the Apple's runtime when compiling for Apple platforms, and the `libBlocksRuntime` from `compiler-rt` on other platforms.
- `objc-sys` chooses the Apple runtime when compiling for Apple platforms, and forces the user to choose a runtime on other platforms.

It definitely makes sense for these to choose the Apple runtime when building for Apple platforms, but we should probably find a better solution for other runtimes (so that the two `-sys` libraries are at least consistent). I see two possibilities:

1. No default, user must choose for the crate to compile.
    - Fewer weird bugs because the wrong runtime was selected.
2. Default `gnustep-1.7` on Linux, `winobjc` on Windows.
    - Easier if you just want to do a quick test (and don't care about the differences in runtime).
    - Tools like [`crater`](https://github.com/rust-lang/crater) can do `cargo check` and `cargo build` (but probably not link).

We should also take into account that these alternative runtimes are probably not used very often, they're mostly useful for internal testing and verification, so maybe letting our crates participate in `crater` runs would be a bigger win?

madsmtm:
Change how the Objective-C runtime is selected by making it a feature on all crates (instead of making it a feature in `-sys` crates, and then "communicating" upwards via. build scripts which runtime was selected).

Also make Apple's runtime the default everywhere, which fixes #140. This means that to use these crates on other systems, you'll have to specify so manually in your `Cargo.toml`, and every dependency has to do so as well. This is a good thing though! It means that e.g. `winit` wouldn't try to "automatically" work on GNUStep, without the `winit` project actually supporting that platform.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,15 +79,13 @@ jobs:
           - name: Test GNUStep
             os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-            host-args: --no-default-features --features gnustep-1-9
-            args: --no-default-features --features objc2-foundation/block --features gnustep-1-9
+            runtime: gnustep-1-9
           - name: Test GNUStep 32bit
             os: ubuntu-latest
             target: i686-unknown-linux-gnu
             cflags: -m32
             configureflags: --target=x86-pc-linux-gnu
-            host-args: --no-default-features --features gnustep-1-9
-            args: --no-default-features --features objc2-foundation/block --features gnustep-1-9
+            runtime: gnustep-1-9
           - name: Test iOS simulator x86 64bit
             os: macos-11
             target: x86_64-apple-ios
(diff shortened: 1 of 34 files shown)


.github/workflows/ci.yml
```diff
@@ -345,26 +345,24 @@ jobs:
 
         cargo run ${{ env.HOSTARGS }} --features assembly --target=$HOST_TARGET test_assembly ${{ env.ARGS }}
 
-    - name: Run Cargo Dinghy
+    - name: Launch XCode Simulator
```
simlay, line 348: I've thought about using something like [futureware-tech/simulator-action](https://github.com/marketplace/actions/launch-ios-simulator) for this type of thing. I'm on the fence about it because the few lines of bash is easier to run locally when things break.

PR #150, opened 2022-05-24
Change how mutability in `msg_send!` is done

madsmtm:
Resolves upstream issue https://github.com/SSheldon/rust-objc/issues/112.

Whichever solution we choose has quite the large effect on https://github.com/madsmtm/objc2/pull/120, since we'd like `msg_send!` and `msg_send_id!` to be consistent, and e.g. `msg_send_id!` sometimes needs the ability to consume its arguments.

This PR is a stab at implementing idea 3 as outlined in https://github.com/SSheldon/rust-objc/issues/112#issuecomment-1134422347. Not quite happy with the fact that we use fully qualified method calls, which means reborrows are sometimes necessary (see diff in `objc2-foundation/src/array.rs` and `objc2-foundation/src/data.rs`, may or may not be a big problem in practice?).

I'm reluctant to do idea 2 (add `msg_send_mut!`) since we would also need `msg_send_bool_mut!` and `msg_send_id_mut!` which is quite cumbersome!

### DIFF
--- a/objc2-foundation/examples/custom_class.rs
+++ b/objc2-foundation/examples/custom_class.rs
@@ -70,12 +70,12 @@ fn main() {
 
     obj.set_number(7);
     println!("Number: {}", unsafe {
-        let number: u32 = msg_send![obj, number];
+        let number: u32 = msg_send![&obj, number];
         number
     });
 
     unsafe {
-        let _: () = msg_send![obj, setNumber: 12u32];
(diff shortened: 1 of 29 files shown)


madsmtm:
I'll go with idea 3, to make `msg_send!` consume its receiver (and require `&[mut] obj` when `obj` is an `Id`), since that is how `msg_send_id!` in #120 will end up working, and then they'll be fully compatible.

PR #172, opened 2022-06-15
Add `rc::Allocated`

madsmtm:
~Blocked on https://github.com/madsmtm/objc2/pull/120.~

Add a new user facing opaque struct `rc::Allocated<T>`, which is intended to be used inside `Id` as `Id<Allocated<T>, O>`, and can be gotten from `msg_send_id![cls, alloc]`. This will help users differentiate between allocated and initialized objects.

See also https://github.com/madsmtm/objc2/issues/87 for more discussion about why we want this.

TODO:
- [x] Documentation
- [x] ~More functionality on `Allocated<T>`?~ Postponed
- [x] ~More functionality on `Id<Allocated<T>, O>`?~ Postponed
- [x] ~Should it be possible to retrieve ivars and initialize them?~ Postponed to #30
- [x] Consider `Allocated<T>(*mut T)`

### DIFF
--- a/objc2/CHANGELOG.md
+++ b/objc2/CHANGELOG.md
@@ -35,6 +35,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   through `&Object`.
 * Added `VerificationError` as more specific return type from
   `Class::verify_sel`.
+* Added `rc::Allocated` struct which is used within `msg_send_id!`.
 
 ### Changed
 * **BREAKING**: `Sel` is now required to be non-null, which means that you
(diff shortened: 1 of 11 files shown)


madsmtm:
We may want to just make it `Allocated<T: Message>(*const T)`, since:
- The ownership doesn't really matter, you shouldn't deref it anyhow
- Deallocation is less efficient if it has to unwrap an `Option` first (very slight issue)
- Whether we should dealloc after alloc on panic may be runtime specific (TODO: Investigate), and this would allow specialized behaviour for that.

madsmtm:
I'll go with `Id<Allocated<T>, O>` over `Allocated<T>(*mut T)` simply because the former requires less code on our end - it's really not that important, both APIs would work fine.

PR #188, opened 2022-07-05
`extern_class!` macro

madsmtm:
Taken out of https://github.com/madsmtm/objc2/pull/161.

Rename `object!` macro to `extern_class!`, use `derive` directly for things that can be derived (instead of us having to implement `PartialEq` and so on manually), and expose the macro publicly.


madsmtm:
See also [prior art in `fruity`](https://github.com/nvzqz/fruity/blob/v0.3.0/src/core/macros.rs)

PR #197, opened 2022-07-16
Add `Id::downgrade`

madsmtm:
Solves remaining part of #58. Unsure about the exact design!

TODO:
- [ ] Limit when not `'static`
- [ ] Maybe add extra trait?
- [ ] Make sure this follows SB (so if we use `Deref`, don't actually go through said deref, instead just use a `cast` directly)
- [ ] Documentation
- [ ] ...


madsmtm:
Superceded by https://github.com/madsmtm/objc2/pull/234

PR #200, opened 2022-07-16
Deprecate using `msg_send!` without comma between arguments

madsmtm:
This is done since the way the macro works currently is not actually valid [according to the reference](https://doc.rust-lang.org/reference/macros-by-example.html#follow-set-ambiguity-restrictions); in particular, an `expr` cannot come after an `ident`, see [this playground link for details](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4b46d5fea92dfe69fafc0a3f21fd4c31).

Also, the previous syntax wasn't even that consistent with Objective-C, we already had to put a comma after the receiver!

I may hold off on merging this until having attained wider adoption, since it is kinda unnecessary for users to deal with as the first thing when they upgrade...


madsmtm:
I've put this behind a feature flag instead as a sort of "soft deprecation", to hopefully enable easier migration

PR #213, opened 2022-07-24
Expose generic version of extern_class!

madsmtm:
I would like to move the macro to `objc2`, and since we use this functionality in `objc2-foundation`, we need to publicly expose it


madsmtm:
Actually, I've changed my mind, it's a hassle to move it to `objc2`

PR #217, opened 2022-07-24
Add `extern_methods!` macro

madsmtm:
Replaces https://github.com/madsmtm/objc2/pull/161.

Syntax:
```rust
extern_methods!(
    // `unsafe` here signifies that 
    unsafe impl NSData {
        #[sel_id(new)] // TODO: Doesn't work yet!
        pub fn new() -> Id<Self, Shared>;

        #[sel(length)]
        pub fn len(&self) -> usize;

        // This has a body, and hence is not touched at all
        pub fn is_empty(&self) -> bool {
            self.len() == 0
        }

        #[sel(bytes)]
        fn bytes_raw(&self) -> *const c_void;
    }
);

// Becomes:

impl NSData {
    pub fn new() -> Id<Self, Shared> {
        unsafe { msg_send_id![Self::class(), new].expect("unexpected NULL NSData") }
    }

    pub fn len(&self) -> usize {
        unsafe { msg_send![self, length] }
    }

    pub fn is_empty(&self) -> bool {
        self.len() == 0
    }

    #[sel(bytes)]
    fn bytes_raw(&self) -> *const c_void;
}
```

TODO:
- [x] Documentation
- [x] Changelog
- [x] ~Find a way to allow using `msg_send_id` and `msg_send_bool` - should be done in a way that `declare_class!` also supports such things!~
  - Postponed. See #239 for the `msg_send_bool` situation. Will be making a `#[sel_id(...)]` attribute, but that is much easier to do with this merged

### DIFF
--- a/objc2/CHANGELOG.md
+++ b/objc2/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   means you'll have to `use objc2::ClassType` whenever you want to use e.g.
   `NSData::class()`.
 * Added `Id::into_superclass`.
+* Added `extern_methods!` macro.
 
 ### Changed
 * **BREAKING**: Change syntax in `extern_class!` macro to be more Rust-like.
--- a/objc2/src/foundation/array.rs
+++ b/objc2/src/foundation/array.rs
@@ -136,34 +136,34 @@ extern_methods!(
     /// Generic accessor methods.
     unsafe impl<T: Message, O: Ownership> NSArray<T, O> {
         #[doc(alias = "count")]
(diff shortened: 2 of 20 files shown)


madsmtm:
I've explicitly used it as:
```rust
extern_methods!(
    ...
);
```

Since that allows rustfmt to format the macro contents. Will document this as well!

madsmtm:
I'm a bit unsure of how big of an improvement this really is. While it _is_ more succinct, error messages get a lot worse, `go to source` doesn't work in docs, and the unsafety is kind of hidden away.

I did find a bug by converting things to use this: `NSArray::objects_in_range` was using `Vec::as_ptr` when it should have been using `Vec::as_mut_ptr`. So that's good, I guess!

PR #222, opened 2022-07-27
Move `objc2-foundation` to `objc2::foundation`

madsmtm:
There really wasn't a good reason that the libraries were separate, they were slowly becoming increasingly linked anyhow, and it really only meant that we were providing a worse user-experience (e.g. docs and examples were scattered, discoverability, versioning).

Quite a major downside is that our version is now tied to `objc2::foundation`, which I expect is going to be making more breaking changes than `objc2` - however, that would really be the case anyway, since `objc2_foundation` would be used basically everywhere that `objc2` is!


madsmtm:
Doc tests don't work ATM, will fix those in a follow-up PR (to make sure that `git history`/`git blame` keeps working)

PR #226, opened 2022-07-29
Improve `NSValue` and add `NSNumber`

madsmtm:
- Remove generics from `NSValue`
- Add safe accessor functions to `NSValue`
- Add `NSNumber`


objc2/src/foundation/geometry.rs
```diff
@@ -25,9 +25,9 @@ pub type CGFloat = InnerFloat;
     not(all(target_os = "macos", target_pointer_width = "32"))
 ))]
 mod names {
-    pub(super) const POINT: &str = "_CGPoint";
-    pub(super) const SIZE: &str = "_CGSize";
-    pub(super) const RECT: &str = "_CGRect";
+    pub(super) const POINT: &str = "CGPoint";
+    pub(super) const SIZE: &str = "CGSize";
+    pub(super) const RECT: &str = "CGRect";
```
madsmtm, line 30: A bit unsure about this, but it seems like it is what `NSValue::encoding` returns when created from one of these

PR #239, opened 2022-08-13
Proper `bool` handling

madsmtm:
4 real this time: Automatic conversion of `bool` to/from the Objective-C `BOOL` across message sends.

Yes, it's unclean, and yes, it couples `objc2-encode` to the runtime (which it will have to be anyhow to handle bitfields properly), and feels like a lot of work for a little thing, but it's simply just a _huge_ productivity boost to never again have to worry about `bool` being special!

Unblocks #217, with this it suddenly becomes much more useful!

Original idea in https://github.com/madsmtm/objc2/pull/126, dismissed back then because of complexity.

Random notes to myself:
- [x] Initial implementation
- [x] Should not be allowed in `extern "C" fn(bool)` in `add_method` (`EncodeArguments`)
- [x] Document special handling of `bool` in `Ivar`
- [x] Document in general
- [x] Figure out how to `cfg` guard the stuff in `objc2-encode` and `objc-sys` properly


ArturKovacs:
This is awesome!

PR #240, opened 2022-08-13
Remove lifetime from `Encoding`

madsmtm:
Seems to make things easier to read without any real downsides, since encodings are almost always statically constructed anyway.

Still unsure if I want to do this though.


madsmtm:
I'm pretty sure I want to add some way to parse encodings after https://github.com/madsmtm/objc2/pull/198, so I think this is a good idea!

PR #244, opened 2022-08-15
Allow returning `Id` from `extern_methods!`

madsmtm:
Initial example syntax:
```rust
extern_methods!(
    unsafe impl MyObject {
        #[sel_id(init)]
        fn new() -> Id<Self, Owned>;

        #[sel_id(description)]
        fn get(&self) -> Id<NSString, Owned>;
    }
);

// Becomes
impl MyObject {
    fn new() -> Id<Self, Owned> {
        // Automatic allocation?
        unsafe { msg_send_id![msg_send_id![Self::class(), alloc], init] }
    }

    fn get(&self) -> Id<NSString, Owned> {
        unsafe { msg_send_id![self, description] }
    }
}
```


madsmtm:
I was previously considering allowing `init` methods to omit the `self`/`this` parameter, but I'll probably defer this for now, given that it is a bit of trouble to implement.

PR #245, opened 2022-08-17
Add `NSSet` and `NSMutableSet`

zatchl:
`NSSet` and `NSMutableSet` have a lot in common with `NSArray` and `NSDictionary`, so they were relatively straightforward to add.

Part of #67.

### DIFF
--- a/objc2/src/foundation/mutable_set.rs
+++ b/objc2/src/foundation/mutable_set.rs
@@ -1,0 +1,368 @@
+use alloc::vec::Vec;
+use core::fmt;
+use core::marker::PhantomData;
+
+use super::set::with_objects;
+use super::{NSCopying, NSFastEnumeration, NSFastEnumerator, NSMutableCopying, NSObject, NSSet};
+use crate::rc::{DefaultId, Id, Owned, Ownership, Shared, SliceId};
+use crate::{ClassType, Message, __inner_extern_class, extern_methods, msg_send_id};
+
+__inner_extern_class!(
+    /// A growable unordered collection of unique objects.
+    ///
+    /// See the documentation for [`NSSet`] and/or [Apple's
+    /// documentation][apple-doc] for more information.
+    ///
+    /// [apple-doc]: https://developer.apple.com/documentation/foundation/nsmutableset?language=objc
+    #[derive(PartialEq, Eq, Hash)]
+    pub struct NSMutableSet<T: Message, O: Ownership = Owned> {
+        p: PhantomData<*mut ()>,
+    }
+
+    unsafe impl<T: Message, O: Ownership> ClassType for NSMutableSet<T, O> {
+        #[inherits(NSObject)]
+        type Super = NSSet<T, O>;
+    }
+);
+
+// SAFETY: Same as NSSet<T, O>
+unsafe impl<T: Message + Sync + Send> Sync for NSMutableSet<T, Shared> {}
+unsafe impl<T: Message + Sync + Send> Send for NSMutableSet<T, Shared> {}
+unsafe impl<T: Message + Sync> Sync for NSMutableSet<T, Owned> {}
+unsafe impl<T: Message + Send> Send for NSMutableSet<T, Owned> {}
+
+extern_methods!(
+    unsafe impl<T: Message, O: Ownership> NSMutableSet<T, O> {
+        /// Creates an empty [`NSMutableSet`].
+        ///
+        /// # Examples
+        ///
+        /// ```
+        /// use objc2::foundation::{NSMutableSet, NSString};
+        /// # #[cfg(feature = "gnustep-1-7")]
+        /// # unsafe { objc2::__gnustep_hack::get_class_to_force_linkage() };
+        ///
+        /// let set = NSMutableSet::<NSString>::new();
+        /// ```
+        pub fn new() -> Id<Self, Owned> {
+            // SAFETY:
+            // Same as `NSSet::new`, except mutable sets are always unique.
+            unsafe { msg_send_id![Self::class(), new] }
+        }
+
+        /// Creates an [`NSMutableSet`] from a vector.
+        ///
+        /// # Examples
+        ///
+        /// ```
+        /// use objc2::foundation::{NSMutableSet, NSString};
+        /// # #[cfg(feature = "gnustep-1-7")]
+        /// # unsafe { objc2::__gnustep_hack::get_class_to_force_linkage() };
+        ///
+        /// let strs = ["one", "two", "three"].map(NSString::from_str).to_vec();
+        /// let set = NSMutableSet::from_vec(strs);
+        /// ```
+        pub fn from_vec(vec: Vec<Id<T, O>>) -> Id<Self, Owned> {
+            // SAFETY:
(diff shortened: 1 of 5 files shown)


madsmtm (commented):
Wow, this looks really nice!

I don't have time for it right now, but will review it in about a week

objc2/src/foundation/mutable_set.rs
```diff
@@ -0,0 +1,291 @@
...
+extern_methods!(
+    unsafe impl<T: Message> NSMutableSet<T> {
+        /// Creates an empty `NSMutableSet`.
+        ///
+        /// # Examples
+        ///
+        /// ```
+        /// use objc2::foundation::{NSMutableSet, NSString};
+        ///
+        /// let set = NSMutableSet::<NSString>::new();
```
madsmtm, line 41: Tip: To get doctests to work on GNUStep:
```suggestion
        /// use objc2::foundation::{NSMutableSet, NSString};
        /// # #[cfg(feature = "gnustep-1-7")]
        /// # unsafe { objc2::__gnustep_hack::get_class_to_force_linkage() };
        ///
        /// let set = NSMutableSet::<NSString>::new();
```

objc2/src/foundation/mutable_set.rs
```diff
@@ -0,0 +1,303 @@
...
+        /// # unsafe { objc2::__gnustep_hack::get_class_to_force_linkage() };
+        ///
+        /// let strs = ["one", "two", "three"].map(NSString::from_str).to_vec();
+        /// let set = NSMutableSet::from_vec(strs);
+        /// ```
+        pub fn from_vec<O: Ownership>(vec: Vec<Id<T, O>>) -> Id<Self, Owned> {
+            // SAFETY:
+            // We always return `Id<NSMutableSet<T, Shared>, Owned>` because
+            // mutable sets are always unique and allow adding/removing elements
+            // but prevent modifying elements.
```
madsmtm, line 67: I'm unsure this is the right tradeoff with mutable sets (also the main reason why `NSArray` and `NSMutableArray` have an ownership parameter; I'm unsure what the right tradeoff is!)

While it makes sense for e.g. `NSString` (which is always `Shared`), I can easily imagine counter-examples where users want to store their elements in a set, and mutably retrieve them afterwards.

madsmtm (changes requested):
Turns out, I got a bit of time to do it now: Looks really nice, especially cool with the detailed examples!

objc2/src/foundation/mutable_set.rs
zatchl: Thanks! I had no idea why doctests were failing.

objc2/src/foundation/mutable_set.rs
zatchl: My thinking was that if users were able to modify the elements (without the set knowing about it) it would invalidate the set. I think that's why Rust's `HashSet`, unlike `Vec`, only has `get` and not `get_mut`.

According to [Apple's NSSet documentation](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Collections/Articles/Sets.html#//apple_ref/doc/uid/20000136-SW1):

> If mutable objects are stored in a set, either the hash method of the objects shouldn’t depend on the internal state of the mutable objects or the mutable objects shouldn’t be modified while they’re in the set.

It's possible for us to prevent users from modifying objects in the set but I don't know how we'd make sure the hash method of mutable objects doesn't depend on its internal state. Maybe it would be enough to document these requirements in `NSSet::get_mut` because I do agree it could be useful.

madsmtm: Oh, right... Duh, forgot how sets work for a moment!

Though an `Id<NSSet, Owned>` would have the property that you could get mutable objects _back_:
```rust
let objs: Vec<Id<NSMutableString, Owned>> = ...; // A list of mutable strings
let set: Id<NSSet<NSMutableString, Owned>, Owned> = NSSet::from_vec(objs); // Deduplicate strings
let objs: Vec<Id<NSMutableString, Owned>> = set.into_vec(); // Get mutable/owned strings back
```

zatchl: That's a good point about getting mutable objects back through `NSSet::into_vec`. I'll give `NSSet` an ownership parameter to make it more like `NSArray`.

objc2/src/foundation/mutable_set.rs
```diff
@@ -0,0 +1,362 @@
...
+        #[doc(alias = "addObject:")]
+        pub fn insert(&mut self, value: Id<T, O>) -> bool {
+            // SAFETY:
+            // We take `Id<T, O>` instead of `&T` because `&T` could be a
+            // reference to an owned object which would cause us to have a copy
+            // of an owned object in our set. By taking `Id<T, O>`, we force the
+            // caller to transfer ownership of the value to us, making it safe
+            // to insert the owned object into the set.
+            let contains_value = self.contains(&value);
+            unsafe { msg_send![self, addObject: &*value] }
```
madsmtm, line 168: It surprised me that this line works, but I guess it does because of some inference rules regarding block statements. I would like to avoid it though, because it seems brittle; for comparison, the following wouldn't work:
```rust
unsafe {
    let contains_value = self.contains(&value);
    msg_send![self, addObject: &*value] // Wouldn't work neither with or without trailing semicolon
    !contains_value
}
```

So please use either `let _: () = ...` or `#[sel(addObject:)] fn add_object(&mut self, &T);`.

Similar for `remove` below

madsmtm (changes requested):
Thanks, just one more thing then I think it's good to go

objc2/src/foundation/set.rs
```diff
@@ -0,0 +1,575 @@
...
+        /// ```
+        /// use objc2::foundation::{NSSet, NSString};
+        ///
+        /// let strs = ["one", "two", "three"].map(NSString::from_str);
+        /// let set = NSSet::from_slice(&strs);
+        /// assert!(set.contains(&NSString::from_str("one")));
+        /// ```
+        #[doc(alias = "containsObject:")]
+        #[sel(containsObject:)]
+        pub fn contains(&self, value: &T) -> bool;
```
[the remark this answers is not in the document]
madsmtm: Apologies, I went and merged https://github.com/madsmtm/objc2/pull/239, could I get you to use that again now?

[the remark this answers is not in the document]
zatchl: For sure. That's a nice improvement being able to use `bool` directly.

objc2/src/foundation/mutable_set.rs
zatchl: I ended up going with the second option of `#[sel(addObject:)] fn add_object(&mut self, &T);` and `#[sel(removeObject:)] fn remove_object(&mut self, &T);`

zatchl:
I just noticed this change 1b323858a1fad85c6e28d7a1fa5e13f7a10ebdfe for `NSArray`. `NSSet` has a similar test and I was wondering if I should make the same change.

madsmtm:
Yeah, `NSSet` seems to behave similarly on my old iPad, it would be nice if you did the same as in that commit

madsmtm (approved):
Thanks again for your work and quick responses! (I'll merge as soon as CI finishes)

PR #246, opened 2022-08-19
Update UI tests for Rust 1.65 nightly

zatchl:
Rust 1.65 seems to have updated their error messages for missing trait implementations which caused some UI tests to fail.


madsmtm:
Thanks!

PR #249, opened 2022-08-29
Add `NSMutableDictionary`

zatchl:
It makes sense to add `NSMutableDictionary` to go with `NSDictionary`.

Part of #67.


objc2/src/foundation/mutable_dictionary.rs
```diff
@@ -0,0 +1,408 @@
...
+            .is_none());
+        assert!(dict
+            .insert(NSString::from_str("two"), NSObject::new())
+            .is_none());
+        assert!(dict
+            .insert(NSString::from_str("three"), NSObject::new())
+            .is_none());
+        assert!(dict
+            .insert(NSString::from_str("one"), NSObject::new())
+            .is_some());
```
madsmtm, line 331: Nit: Formatting seems wrong here?

zatchl: It definitely looks weird, but that's how `rustfmt` decided to format each `assert!`. I changed the dictionary to use `NSNumber` for the keys so that each `assert!` is short enough to fit on a single line.

zatchl:
I'm planning to use this library to wrap CoreWLAN and eventually write a cross-platform Wi-Fi library. CoreWLAN returns Wi-Fi scan results in an `NSSet`, which explains why I contributed that first. I don't think I'll personally need `NSMutableDictionary`, but I just find adding functionality from Foundation to be interesting and hopefully useful to others.

madsmtm:
Oooh, interesting! Do drop me a note somewhere when you have something to show, I would be glad to give it a soundness review.

zatchl:
Definitely! That would be great.

PR #264, opened 2022-09-08
Automatic binding generation

madsmtm:
This is very similar to [using `bindgen`](https://github.com/madsmtm/objc2/issues/85), but I went with creating things from scratch because:
1. `bindgen` is kinda intimidating, knowing so little about `libclang`, so I started here first; I'd still like to upstream some of this, but I'll start here for now.
2. This only has to run on my machine every once in a while, whereas `bindgen` has to handle all C code under the sun. A few manual fixes and such are acceptable.
3. I'd like to run this multiple times for different targets (iOS, macOS, tvOS, ... - GNUStep will probably suffer a bit here), and then merge the result using `#[cfg]` so that we can ship one, ready-to-use library.
4. I'd like to enrich the output using separate files (`.apinotes` or similar), and such a thing will probably never belong in `bindgen`
5. Other things that are harder to integrate into something existing, for example availability attributes.

That said, I _have_ taken inspiration from the existing Objective-C implementation in `bindgen`, credits to @simlay.

In the end, I'm envisioning something like this (so `objc2` would be the Apple equivalent to the `windows` crate):
```
objc2/src/
  foundation/
    generated/
      ...
    unsafe_fns.toml // Or `Foundation.apinotes`, or similar
    string.rs
    mutable_string.rs
    ...
  appkit/...
  core_data/...
  core_display/...
  // And so on, for the Apple frameworks that expose an Objective-C interface
```

See https://github.com/madsmtm/objc2/issues/85#issuecomment-1236222148 for a bit of details on what Swift does.


madsmtm:
I estimated the size of the code if we included _all_ of Apple's frameworks (assuming Rust and C have similar line ratios); it would be around 13MB zipped, 50MB unzipped. For reference, the `windows` crate is 3MB zipped, 35MB unzipped.

madsmtm:
I probably won't put this under `objc2`, instead I'll create a new crate with the frameworks. Possible names, in order of my current preference:
1. `apple`, is an empty crate owned by @zu1kd, have sent them an email to discuss name transfer
2. `icrate`, a play on Apple's "i"-prefixing (iPhone/iPad/...)
3. `ilib`, similar to above
4. `frameworks`, ambiguous since also commonly used for "web frameworks"
5. `appel`, wordplay

madsmtm:
Example of how I want at least some of AppKit to look like: https://github.com/rust-windowing/winit/tree/fafdedfb7d3a7370ca4b01108f7713b685633164/src/platform_impl/macos/appkit

madsmtm:
`init` methods always return the same type as the receiver: https://clang.llvm.org/docs/AutomaticReferenceCounting.html#related-result-types

How can we model this nicely?

madsmtm:
Idea: use Swifty names? https://github.com/apple/swift-evolution/blob/main/proposals/0005-objective-c-name-translation.md

madsmtm:
Idea: Use (parts of) Swift's test suite (for example this one for errors: https://github.com/apple/swift/blob/7123d2614b5f222d03b3762cb110d27a9dd98e24/test/Inputs/clang-importer-sdk/usr/include/errors.h)

madsmtm:
Clang has ["module maps"](https://clang.llvm.org/docs/Modules.html#module-map-language) for figuring out what to import from Objective-C headers - we should use those!

ericmarkmartin:
Hey,

I'm really excited about this enhancement and would find it super useful for some work I'm doing on binding the `AuthenticationServices` framework. That in mind, I have a couple of high-level questions.

1. do you have a rough ETA for when this thing might land? If it's soon-ish I'll probably hold off on writing my own bindings but otherwise I might just proceed on my own.
2. if you think this is still a little ways out, is there any way I could help to bring this around sooner?

madsmtm:
Currently I'm trying to push forwards on an initial version that actually compiles `Foundation` and `AppKit`.

Primary missing items are:
- Proper imports
- Struct definitions
- A few remaining typedefs
- https://github.com/madsmtm/objc2/pull/244
- The remaining stuff to get things to compile

Timeline on this is unknown, but I have quite a lot of free time this week, so I should be able to get pretty far. I can probably give you a better estimate at the end of the week, if I'm not done by then.

Once I have this finished, there'll be lots of smaller things to work on, which I could probably use some help with, including:
- Actual type-safety (creating newtypes instead of type-aliases for `enum`s, `typedef`s and so on, allowing us to later on mark various methods as safe)
- Extern functions
- Inline functions (requires quite a lot, since we effectively have to translate C code to Rust. But we can probably get quite far with a few tricks)
- Block support
- Protocols: #250
- "out" parameters: #277
- Autogenerated documentation (using code comments, and ideally data from `developer.apple.com`)
- Possibly a more "swifty" translation (tweaking method names, making enums associated constants). Probably better to do after we have autogenerated documentation, since there is great value in the current design where you can very easily find what you're looking for

I'll try to see if I can figure out a good way for you / somebody else to contribute with some of that, when I get there.

madsmtm:
I could use some guidance though: I'll admit I'm not the best at `git`, and I'm unsure if storing the generated result in `git` is the best solution? It does make audits (security and just general) easier, at the cost of blowing up the amount files we store. An alternative could also be to create a new repo just for it.

But what do you think?

header-translator/src/main.rs
```diff
@@ -0,0 +1,311 @@
...
+            fs::write(&path, output).unwrap();
+
+            (name, declared_types)
+        })
+        .collect();
+
+    let mut tokens = String::new();
+
+    for (name, _) in &declared {
+        writeln!(tokens, "pub(crate) mod {name};")?;
```
simlay, line 285: Have you thought about using [`quote`](https://crates.io/crates/quote)? It's been my experience that it helps when actually generating the rust to verify correctness.

header-translator/src/main.rs
```diff
@@ -0,0 +1,311 @@
...
+
+                    f(&library, &name, entity);
+                }
+            }
+        }
+        EntityVisitResult::Continue
+    });
+}
+
+fn output_files(
```
simlay, line 257: Should you want to use this crate as a [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html), I think `output_files` (and other stuff) will need to go in the `lib.rs`

simlay (commented):
> I could use some guidance though: I'll admit I'm not the best at `git`, and I'm unsure if storing the generated result in `git` is the best solution? It does make audits (security and just general) easier, at the cost of blowing up the amount files we store. An alternative could also be to create a new repo just for it.
> 
> But what do you think?

Putting it in a different repo isn't a bad idea but I think in general what I'd say is that the generated stuff in tracking should be either for some kind of sys-crate (or `icrate`) or for integration tests. [Bindgen keeps the input headers and the expectations in git](https://github.com/rust-lang/rust-bindgen/tree/master/bindgen-tests/tests/expectations/tests) and verifies that they match as expected in CI and the output does compile. Using the full frameworks to test is useful to know breadth of functionality but I believe it can change from macOS/SDK version to version.

After reviewing some stuff, it seems like the `header-translator` crate is meant to run as a binary and then either store the output in the crate and tracking. I tend to like using a generator crate in a [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html).

header-translator/src/main.rs
madsmtm: Yeah, I started using that, but found that it was difficult to generate "good looking" Rust code (`rustfmt` can only go so far when things are nested in macros), and it wasn't really satisfactory wrt. performance.

madsmtm:
> I tend to like using a generator crate in a build script

Yeah, I'm aware of that approach, my desire with `header-translator`/`icrate` was to do things a bit differently than `bindgen`, to improve:
- Compilation speed, since the work is only done once on my local machine, and not on every user's machine (here I'm hoping that download speed will not be a limiting factor)
- Documentation, since items that are only available on one platform would be hidden behind appropriate `cfg`s
- The need to have a specific SDK version installed (would help with cross-compiling on Linux)

> I believe it can change from macOS/SDK version to version.

I think my desire was to manually keep `icrate` up to date with new SDK versions. I'm currently compiling for the macOS 12.3 SDK, on a machine running macOS 10.14.6.

madsmtm:
Status: `cargo build -picrate --features Foundation` now works.

A clean build takes ~25 seconds, a build with a single file changed takes ~11 seconds (on my pretty old machine). 

Using `cargo +nightly rustc -picrate --features Foundation -- -Z self-profile && summarize summarize icrate-{pid}.mm_profdata`, we can get rustc to tell us where it's spending its time.
Clean build:
```
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| Item                                                | Self time | % of total time | Time     | Item count | Incremental result hashing time |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| LLVM_module_codegen_emit_obj                        | 19.54s    | 35.855          | 19.54s   | 256        | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| LLVM_passes                                         | 6.88s     | 12.623          | 6.88s    | 1          | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| codegen_module                                      | 3.91s     | 7.181           | 4.64s    | 256        | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| expand_crate                                        | 3.86s     | 7.072           | 3.86s    | 1          | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| codegen_crate                                       | 2.58s     | 4.728           | 7.24s    | 1          | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| typeck                                              | 1.99s     | 3.658           | 2.27s    | 15323      | 59.69ms                         |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| mir_borrowck                                        | 1.71s     | 3.144           | 3.99s    | 15323      | 7.70ms                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| mir_drops_elaborated_and_const_checked              | 743.11ms  | 1.363           | 888.30ms | 15323      | 4.41ms                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| incr_comp_intern_dep_graph_node                     | 654.54ms  | 1.201           | 1.14s    | 1556795    | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| LLVM_module_optimize                                | 640.50ms  | 1.175           | 640.50ms | 256        | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| mir_built                                           | 627.18ms  | 1.151           | 1.06s    | 15323      | 162.19ms                        |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| LLVM_module_codegen                                 | 569.89ms  | 1.045           | 20.11s   | 256        | 0.00ns                          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
// SNIP
```

Single file changed:
```
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| Item                                             | Self time | % of total time | Time     | Item count | Incremental load time | Incremental result hashing time |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| expand_crate                                     | 4.48s     | 38.074          | 4.48s    | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| generate_crate_metadata                          | 639.36ms  | 5.439           | 1.87s    | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| incr_comp_load_dep_graph                         | 536.81ms  | 4.567           | 536.81ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| hir_crate                                        | 483.00ms  | 4.109           | 582.56ms | 1          | 0.00ns                | 4.18µs                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| incr_comp_encode_dep_graph                       | 376.30ms  | 3.201           | 376.30ms | 1569613    | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| resolve_access_levels                            | 313.73ms  | 2.669           | 313.73ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| link_rlib                                        | 276.83ms  | 2.355           | 276.83ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| incr_comp_query_cache_promotion                  | 245.21ms  | 2.086           | 352.09ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| encode_query_results_for                         | 243.17ms  | 2.069           | 243.17ms | 61         | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| optimized_mir                                    | 242.48ms  | 2.063           | 244.80ms | 80         | 242.43ms              | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| monomorphization_collector_graph_walk            | 224.45ms  | 1.910           | 523.33ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| late_resolve_crate                               | 216.57ms  | 1.843           | 216.57ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| incr_comp_serialize_result_cache                 | 176.08ms  | 1.498           | 419.39ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| privacy_access_levels                            | 157.35ms  | 1.339           | 248.73ms | 1          | 0.00ns                | 3.16ms                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| codegen_copy_artifacts_from_incr_cache           | 122.74ms  | 1.044           | 122.74ms | 256        | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
| crate_lints                                      | 118.53ms  | 1.008           | 296.36ms | 1          | 0.00ns                | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+-----------------------+---------------------------------+
// SNIP
```

I would like to improve this, but don't really know where I should look. Perhaps I'll try to see if putting everything in one file is better? Otherwise, perhaps `extern_methods!` can be made more inefficient?

madsmtm:
Putting everything in one file does not seem to affect the result in any meaningful way, so doing the imports in a different way also probably won't have any impact.

madsmtm:
For comparison, I tried compiling the following:
```toml
[dependencies.windows]
version = "0.43.0"
features = [
    "Foundation_Collections",
    "Foundation_Diagnostics",
    "Foundation_Metadata",
    "Foundation_Numerics",
    "Data_Xml_Dom",
    "Win32_Foundation",
    "Win32_Security",
    "Win32_System_Threading",
    "Win32_UI_WindowsAndMessaging",
]
```

There, the compiler spent its time as follows:
```
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| Item                                                | Self time | % of total time | Time     | Item count |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| LLVM_module_codegen_emit_obj                        | 9.65s     | 36.293          | 9.65s    | 16         |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| LLVM_passes                                         | 2.74s     | 10.320          | 2.74s    | 1          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| codegen_module                                      | 2.10s     | 7.903           | 2.42s    | 16         |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| typeck                                              | 1.84s     | 6.934           | 2.05s    | 19480      |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| mir_borrowck                                        | 1.44s     | 5.413           | 2.87s    | 19480      |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| mir_drops_elaborated_and_const_checked              | 743.73ms  | 2.798           | 862.04ms | 19480      |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| expand_crate                                        | 525.88ms  | 1.978           | 551.30ms | 1          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| hir_crate                                           | 430.01ms  | 1.618           | 473.48ms | 1          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| eval_to_allocation_raw                              | 347.30ms  | 1.307           | 886.66ms | 13881      |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| mir_built                                           | 336.45ms  | 1.266           | 608.52ms | 19480      |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| check_well_formed                                   | 326.72ms  | 1.229           | 565.58ms | 28414      |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| LLVM_module_optimize                                | 319.08ms  | 1.200           | 319.08ms | 16         |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
| generate_crate_metadata                             | 315.63ms  | 1.187           | 1.76s    | 1          |
+-----------------------------------------------------+-----------+-----------------+----------+------------+
// SNIP
```

Notably, we spend more time in the `expand_crate` step (since the `windows` crate has expanded most macros by default), but percentage-wise everything else looks roughly the same.

Thinking about it, it _does_ make sense that codegen'ing `icrate` takes a long time, every method we have there requires creating new static for the selector along with a generic call to `MessageReceiver::send_message` that needs to be fully instantiated!

madsmtm:
Note to self: Swift has a pretty comprehensive test suite for their Objective-C import functionality, maybe we could reuse a few headers from there? https://github.com/apple/swift/blob/7123d2614b5f222d03b3762cb110d27a9dd98e24/test/Inputs/clang-importer-sdk/usr/include/errors.h

madsmtm:
Wrt. compilation times: I'm considering adding features such as `Foundation_NSArray` to enable functionality from `Foundation/NSArray.h` (`NSArray`, `NSMutableArray`, and probably all the things that are transitively enabled by this, such as `NSString`, `NSObject`, `NSOrderedCollectionDifferenceCalculationOptions`, ...). Of course the `Foundation` feature would still just enable everything.

madsmtm:
`AppKit` now compiles!

Clean build 1m 30s, clean check build 52s.

ericmarkmartin:
Do you have instructions somewhere on what you're doing to build appkit? I'd like to try building AuuthenticationServices

madsmtm:
> Do you have instructions somewhere on what you're doing to build appkit? I'd like to try building AuuthenticationServices

There's a bit of docs in `header-translator/README.md`, but for adding new frameworks you need to do a few extra. I've pushed commit https://github.com/madsmtm/objc2/pull/264/commits/65643ff513adae75d84bf33db33c3ded71753c3e which adds support for `AuthenticationServices`, so you can try that.

As you can see there, I needed to do a few manual fixes to work around e.g. `ASViewController` being a typedef for `NSViewController` or `UIViewController`. Since the different frameworks are enabled with features, I'd have a bit of a hard time enabling `AppKit` or `UIKit` conditionally, so I just stubbed it out for now.

Note also that protocols are not properly supported yet, so it is currently probably currently of limited use.

madsmtm:
I've thought about the `git` situation some more, and I think I'll create a separate repository containing only the generated artifacts, and link it to this one as a `git` submodule. The reasoning is to avoid blowing up the `git` history of this project, as well as to protect the rest of the project against DMCA takedowns in case Apple decide that they don't like autogenerated derivations of their header files.

Speaking of, does anyone here know which license it would be proper to release `icrate` under? The `license` field supports `OR` and `AND`, so we could do something there? Or maybe we should just provide our own `license-file` field, and then specify that the user must comply with Apple's terms there?

madsmtm:
I originally did [Objective-C categories](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocCategories.html) with just a basic `#[doc = "category name"] impl ClassName { ... }`, which was possible because everything lives in the same crate.

However, I'm beginning to doubt that that is wise, partly because it locks us into always distributing everything as one crate, but primarily because if the user needs to enable functionality with cargo features, it can quickly become _very_ difficult to discern when the `-[NSObject validateMenuItem:]` method is available - this is much easier if one has already imported something like `trait NSObject_NSMenuValidation`.

Though it probably still makes sense to do a plain `impl` for things that live in the same file.

simlay:
> However, I'm beginning to doubt that that is wise, partly because it locks us into always distributing everything as one crate, but primarily because if the user needs to enable functionality with cargo features, it can quickly become _very_ difficult to discern when the `-[NSObject validateMenuItem:]` method is available - this is much easier if one has already imported something like `trait NSObject_NSMenuValidation`.

This is kinda how bindgen does it right now. The inheritance part is a bit rough as seen in [ukit-sys `UIButton`](https://simlay.net/uikit-sys/master/uikit_sys/struct.UIButton.html#impl-NSObject_UIAccessibility). There are 50+ traits implemented on the `UIButton` struct.

I've run into two issues/features with this:
* The "apple docs" mention various functions and such but don't say which category it is in so one has to  start guessing which trait a given function/method is in. Honestly, if all the category functions and methods for a given Objective-C class could be in one jumbo-trait, that might make the UX better. I'm sure there's a trade off for this doing that.
* From what I recall (~2 years ago), if `struct A` and `struct B` both implement `trait C` (with say an `fn foo`), if an instance of `A`  and an instance of `B` call `foo` in the same module (maybe namespace is the right word) some extra/uncommon trait notation was needed. I just tried this on an example of uikit-sys and couldn't reproduce this annoyance so I could be remembering wrong or the rust compiler got smarter.

madsmtm:
> This is kinda how bindgen does it right now. The inheritance part is a bit rough as seen in [ukit-sys `UIButton`](https://simlay.net/uikit-sys/master/uikit_sys/struct.UIButton.html#impl-NSObject_UIAccessibility). There are 50+ traits implemented on the `UIButton` struct.

I don't think it would be as bad in our case, since `UIButton` ultimately derefs to `NSObject`, so we don't need to implement these traits for anything else than the actual  ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=929ec4bf760c72c9680e228150f87b2b)).

Note that this won't work for class methods, but I doubt that will be a problem - for `UIButton`, there is basically nothing that is required to be called on the `UIButton` class (e.g. that can't just be called from a superclass like `UIView`, `UIResponder` or `NSObject`).

>     * The "apple docs" mention various functions and such but don't say which category it is in so one has to  start guessing which trait a given function/method is in. Honestly, if all the category functions and methods for a given Objective-C class could be in one jumbo-trait, that might make the UX better. I'm sure there's a trade off for this doing that.

Relatedly, the name is from the Objective-C side not part of the public API, and Apple may decide to change it.

Perhaps I'll try to do a bit of "if defined in the same file" logic, and if not, just ignore those categories for now. Or maybe I'll end up special-casing `NSObject`, since that seems like a really popular candidate for categories.

>     * From what I recall (~2 years ago), if `struct A` and `struct B` both implement `trait C` (with say an `fn foo`), if an instance of `A`  and an instance of `B` call `foo` in the same module (maybe namespace is the right word) some extra/uncommon trait notation was needed. I just tried this on an example of uikit-sys and couldn't reproduce this annoyance so I could be remembering wrong or the rust compiler got smarter.

Perhaps it's if there's two traits `C` and `D` both with an `fn foo`? In those cases you need to use ["fully qualified syntax"](https://doc.rust-lang.org/reference/paths.html#qualified-paths).

notgull:
Well, I was writing an automatic binding generator for AppKit, only to find out that you're doing it already; and from a glance it looks like your version is *way* ahead of mine! :-)

My only complaint would be that, in my opinion, having the setup use `Deref` in order to emulate inheritance isn't hygenic. If it were up to me, I'd do something like what `gtk-rs` does:

```rust
// Have this trait.
pub trait IsA<T> {}

pub struct NSDate { /* .. */ }
impl IsA<NSDate> for NSDate {}
impl IsA<NSObject> for NSObject {}

// Extension trait.
pub trait NSDateExt : IsA<NSDate> { /* ... */ }
impl<T: IsA<NSDate>> NSDateExt for T {}
```

That being said, I'm very excited to see this once it reaches readiness. Please let me know if there's anything I can do for this.

madsmtm:
> My only complaint would be that, in my opinion, having the setup use `Deref` in order to emulate inheritance isn't hygenic. If it were up to me, I'd do something like what `gtk-rs` does:
> 
> ```rust
> // Have this trait.
> pub trait IsA<T> {}
> 
> pub struct NSDate { /* .. */ }
> impl IsA<NSDate> for NSDate {}
> impl IsA<NSObject> for NSObject {}
> 
> // Extension trait.
> pub trait NSDateExt : IsA<NSDate> { /* ... */ }
> impl<T: IsA<NSDate>> NSDateExt for T {}
> ```

Yeah, still very much experimenting with what's the best approach!

I'm not sure what you mean by "isn't hygenic", would you care to elaborate?

> That being said, I'm very excited to see this once it reaches readiness. Please let me know if there's anything I can do for this.

Sure, there will be fairly soon.

I'm done with the rebase and do actually have something workable over in https://github.com/madsmtm/objc2/pull/308 now, once that is merged it should become easier to collaborate.

madsmtm:
A possible source of at least a bit of docs is also https://github.com/xamarin/apple-api-docs.

In general, Xamarin has done a lot of work in this space, I should try to take a look at their stuff at some point

madsmtm:
https://github.com/madsmtm/objc2/pull/308 is done 🎉

Will try to release a new version of `icrate` to `crates.io` today, so that you can play with it. I've opened [tagged issues](https://github.com/madsmtm/objc2/labels/A-icrate) for some of the remaining stuff - do free to comment on any of them if you have suggestions for how to do something differently (especially ergonomics-wise, the hardest part in all of this is in the design work ;) )

PR #276, opened 2022-10-05
Better error handling in `msg_send!` and `msg_send_id!`

madsmtm:
## Background

Many methods take an `NSError**` as their last parameter, which is used to communicate errors to the caller, see [Error Handling Programming Guide For Cocoa](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorHandling/ErrorHandling.html).

Swift [has a convention](https://developer.apple.com/documentation/swift/about-imported-cocoa-error-parameters) for when they convert such methods, we should have something similar.

Note that only methods that return `BOOL` or an instance are supported, because, as per the documentation:
> Success or failure is indicated by the return value of the method. Although Cocoa methods that indirectly return error objects in the Cocoa error domain are guaranteed to return such objects if the method indicates failure by directly returning `nil` or `NO`, you should always check that the return value is `nil` or `NO` before attempting to do anything with the `NSError` object.

Examples of a few different methods in Foundation that return errors:
- [`-[NSDictionary writeToURL:error:]`](https://developer.apple.com/documentation/foundation/nsdictionary/2879139-writetourl?language=objc)
- [`-[NSExtensionContext cancelRequestWithError:]`](https://developer.apple.com/documentation/foundation/nsextensioncontext/1412773-cancelrequestwitherror?language=objc)
- [`-[NSObject validateValue:forKey:error:]`](https://developer.apple.com/documentation/objectivec/nsobject/1416754-validatevalue?language=objc)
- [`-[NSAppleScript compileAndReturnError:]`](https://developer.apple.com/documentation/foundation/nsapplescript/1407582-compileandreturnerror?language=objc)
- [`-[NSNumberFormatter getObjectValue:forString:range:error:]`](https://developer.apple.com/documentation/foundation/nsnumberformatter/1412588-getobjectvalue?language=objc)
- [`+[NSData dataWithContentsOfFile:options:error:]`](https://developer.apple.com/documentation/foundation/nsdata/1547244-datawithcontentsoffile?language=objc)

## Implementation

The idea is that we change `msg_send!` and `msg_send_id!` to support specifying `error: _` as the last part of the selector (`_` is not a valid expresion, so this is easy to distinguish), and if so, performs either a `BOOL != NO` check, or a NULL check, and returns a `Result` with the desired value or the error. Example:

```rust
fn myMethodThatReturnsError(&self, that_returns: i32) -> Result<(), Id<NSError, Shared>> {
    unsafe { msg_send![self, myMethod: that_returns, error: _] }
}
```

This way, it is very easy to specify, while also only happening if the user explicitly requests it!

### DIFF
--- a/objc2/CHANGELOG.md
+++ b/objc2/CHANGELOG.md
@@ -23,7 +23,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
       msg_send_id![NSObject::alloc(), init]
   };
   ```
-* Add `Class::class_method`.
+* Added `Class::class_method`.
+* Added the ability to specify `error: _`, `somethingReturningError: _` and
+  so on at the end of `msg_send!`/`msg_send_id!`, and have it automatically
+  return a `Result<..., Id<NSError, Shared>>`.
+* Added the ability to specify an extra parameter at the end of the selector
+  in methods declared with `extern_methods!`, and let that be the `NSError**`
+  parameter.
 
 ### Changed
 * Allow other types than `&Class` as the receiver in `msg_send_id!` methods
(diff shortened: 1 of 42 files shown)


madsmtm:
I explicitly left the error type generic, since after #264 `NSError` won't be in the `objc2` crate, and then I'd have to redesign things

Issue #285, opened 2022-11-08
Support `cfg` in `declare_class!`


madsmtm:
[Reported on `winit` matrix](https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$FD0lqA7d3quWZK6kk6_CbnkOAhtPMYiLRrXytYwDWb4?via=libera.chat&via=matrix.org&via=safaradeg.net).

Code excerpt from that thread:

```rust
declare_class! {
    // -- snip --
    unsafe impl ApplicationDelegate {
        // -- snip --
        #[cfg(feature = "macos-open-urls")]
        #[sel(application:openURLs:)]
        fn open_urls(&self, _sender: *const Object, urls: &NSArray<NSURL, Shared>) {
            trace_scope!("application:openURLs:");
        }
    }
}
```

madsmtm:
Implementation-wise: `__attribute_helper`'s `@extract_sel` has to be extended to also extract attributes that apply to the function as well as any items referencing the function (which we do internally in code similar to `classbuilder.add_method(ApplicationDelegate::open_urls`).

Current list (based on [the built-in index](https://doc.rust-lang.org/reference/attributes.html#built-in-attributes-index)):
- `cfg`
- `deprecated` - We should add a corresponding `allow(deprecated)` before use
- Any availability attributes that we decide on: https://github.com/madsmtm/objc2/issues/266
- `cfg_attr`, if the inner attributes are on this list. (Might punt on this one, since it's a bit difficult, and a rare case anyhow)

PR #286, opened 2022-11-09
Handle `cfg` in `declare class!`

madsmtm:
Fixes https://github.com/madsmtm/objc2/issues/285.

`cfg_attr` is not yet handled, but that doesn't work properly in `extern_methods` either, so instead I've just documented that that's the case, and encouraged people to open an issue if they have a need for it.


madsmtm:
This also allows implementing a protocol or not based on a `cfg`

PR #287, opened 2022-11-09
Move crates into a subfolder

madsmtm:
It is becoming increasingly clear to me that the number of crates in this repo will basically just continue to grow, so let's move crates into a subfolder to make things easier to navigate.


madsmtm:
I squashed this instead of creating a merge commit, to hopefully make `git blame` easier to work with.

Issue #291, opened 2022-11-16
Improving how we map protocols


madsmtm:
Continuing on the work done in https://github.com/madsmtm/objc2/pull/250.

### Currently

A protocol is mapped to a struct that impls `Message`, which allows using it in `Id`.
This allows similar patterns as `id<ProtocolName>` would do in Objective-C, but has a lot of tradeoffs, including: 

1. Calling a method on a protocol is more cumbersome than doing it on a normal struct.
  
    ```rust
    use icrate::Foundation::{NSLock, NSLocking};
    
    let obj = NSLock::new();
    
    // Not possible to call `lock` automatically, we have to convert it to a protocol object first
    // obj.lock();
    
    let proto: &NSLocking = obj.as_protocol();
    proto.lock();
    proto.unlock();
    ```

2. Protocol class methods are impossible to do
    
    ```rust
    use icrate::Foundation::NSString;
    let res: bool = unsafe { msg_send![NSString::class(), supportsSecureCoding] };
    ```

3. Calling methods on protocols that has multiple parent protocols is cumbersome:
    
    ```rust
    use icrate::AppKit::{NSTextCheckingClient, NSTextInputClient, NSTextInputTraits};
    let proto: &NSTextCheckingClient = ...; // Inherits NSTextInputClient and NSTextInputTraits
    proto.addAnnotations_range(...);

    // proto.hasMarkedText();
    let parent1: &NSTextInputClient = proto.as_protocol();
    parent1.hasMarkedText();

    // proto.autocorrectionType();
    let parent2: &NSTextInputTraits = proto.as_protocol();
    parent2.autocorrectionType();
    ```

### Ideally

Ideally the constraints outlined above would be gone

1.
    ```rust
    use icrate::Foundation::{NSLock, NSLocking};
    
    let obj = NSLock::new();
    
    obj.lock(); // Calls -[NSLocking lock]

    // Still possible
    let proto: &NSLocking = obj.as_protocol();
    
    proto.unlock();
    ```

2.
    ```rust
    use icrate::Foundation::{NSString, NSSecureCoding};
    let res = NSString::supportsSecureCoding();
    ```

3.
    ```rust
    use icrate::AppKit::{NSTextCheckingClient, NSTextInputClient, NSTextInputTraits};
    let proto: &NSTextCheckingClient = ...;
    proto.addAnnotations_range(...);
    proto.hasMarkedText();
    proto.autocorrectionType();
    ```

### Idea

[...]

PR #293, opened 2022-11-23
Better protocols

madsmtm:
Fixes https://github.com/madsmtm/objc2/issues/291.

TODO:
- [x] Finish basic implementation
- [x] Update documentation
- [x] ~Use new implementation for `NSCopying`, `NSMutableCopying`, ...~ Deferred to https://github.com/madsmtm/objc2/issues/401
- [x] Test assembly output to ensure that using `dyn` like this doesn't incur extra overhead
  - It doesn't, see [this godbolt](https://rust.godbolt.org/z/rnGs5YMWb).


madsmtm:
A somewhat end-goal is that we can use this in `icrate` for translating places that take/return something like `id<NSCopying + NSFastEnumeration>` to:
```rust
ProtocolObject<dyn NSCopying + NSFastEnumeration>
```

To use such an object, we would like the following implementations to exist:
```rust
// Required to use the object at all
impl NSCopying for ProtocolObject<dyn NSCopying + NSFastEnumeration> {}
impl NSFastEnumeration for ProtocolObject<dyn NSCopying + NSFastEnumeration> {}

// Would be nice to have
impl ConformsTo<dyn NSCopying> for ProtocolObject<dyn NSCopying + NSFastEnumeration> {}
impl ConformsTo<dyn NSFastEnumeration> for ProtocolObject<dyn NSCopying + NSFastEnumeration> {}

// Not really that important, but could be cool
impl ConformsTo<dyn NSCopying + NSFastEnumeration> for ProtocolObject<dyn NSCopying + NSFastEnumeration> {}
```

---

```rust
// For an object `MyClass` that implements the protocols, these are required
impl NSCopying for MyClass {}
impl NSFastEnumeration for MyClass {}

// And these would be expected
impl ConformsTo<dyn NSCopying> for MyClass {}
impl ConformsTo<dyn NSFastEnumeration> for MyClass {}

// To properly handle the case above, this one would be nice
impl ConformsTo<dyn NSCopying + NSFastEnumeration> for MyClass {}
```

madsmtm:
I've been going about this a bit wrong; what we _want_ is for users to specify this:
```rust
extern_protocol!(
    unsafe trait NSCopying {
        // ...
    }
);

extern_protocol!(
    unsafe trait NSMutableCopying: NSCopying {
        // ...
    }
);

extern_protocol!(
    unsafe trait NSFastEnumeration {
        // ...
    }
);

extern_class!(
    struct MyClass;
);

unsafe impl NSCopying for MyClass {}
unsafe impl NSMutableCopying for MyClass {}
unsafe impl NSFastEnumeration for MyClass {}
```

Notice how the protocol implementations for the class is _not_ in the `extern_class!` macro? That's intentional, since the protocol may be specified in a "later"/"higher up" framework than the class is (which would work for Objectice-C, since they'd just define a category, but it won't work for us).

Additionally, the use of `ConformsTo` is entirely hidden, which is also really nice!

madsmtm:
```rust
extern_protocol!(
    unsafe trait FooBar: Foo + Bar {}
);

// Generates

unsafe trait FooBar: ... {...}

unsafe impl<T: ?Sized + FooBar> FooBar for ProtocolObject<T> {}

// Note: Needs to be done a bit differently because of the orphan rule
impl<T: ?Sized + FooBar> ConformsTo<dyn FooBar> for T {}

// Maybe?
// impl<T: ?Sized + FooBar> ConformsTo<dyn Foo> for T {}
// impl<T: ?Sized + FooBar> ConformsTo<dyn Bar> for T {}
```

Issue #203, opened 2022-07-19
Minimum Supported Rust Version (MSRV)?


madsmtm:
This library should probably settle on some kind of MSRV policy, I would like some input on that?

Notes:
- Commonly used is "Latest - 3"
- GCC Rust will support 1.40 to start with
- `mrustc` supports 1.54 (as of july 2022)
- We could have different policies for the different crates (e.g. `objc-sys` supports "Latest - 10" while `objc2` supports "Latest - 3")
- Downside is that features like `extern "C-unwind"` would take longer to make default

madsmtm:
See also https://github.com/rust-lang/libs-team/issues/72#issuecomment-1192671625 on how to check with `-Zminimal-versions` in CI, if that becomes necessary

Lokathor:
I would vote for 12 months timeframe, no more. I made some strong statements about a fixed MSRV for bytemuck and tinyvec, and it's not impossible to stand by them. but it's inconvenient to stick with them eternally. A rolling window lets you eventually adapt to rust updates without cfg all over your crate, while a large rolling window lets you stay compatible with linux distros.

Even SDL2, famed for its long term compatibility, will bump its requirements over time.

matklad:
> while a large rolling window lets you stay compatible with linux distros.

Something I've learned is that 12 months might not be enough for some LTS-style situations.

* 12 months is too fast for Debian stable
* curiously, tokio has 6 months MSRV, but also 1 year LTS releases, which together effectively gives 18 months MSRV

notgull:
For my crates, I usually try to keep the MSRV at or below Debian Stable, but I don't think that will be desired or even possible for this crate. As it's "experimental" for the time being, I think it'd be fine to have an undefined MSRV. But, eventually, it would be nice to pin it to either a timed MSRV (12 months is usually fine) or some distro's Rust version (Debian Testing uses 1.56 as of the time of writing).

madsmtm:
Any bump will also have to be within Firefox' MSRV, if we want `wgpu` to use `icrate`: https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/util.py

madsmtm:
Some more thoughts were written in https://github.com/madsmtm/objc2/issues/656#issuecomment-2381409778, an important point:

> there are no distros lagging behind and holding back the ecosystem

[...]

Issue #174, opened 2022-06-16
Replace `objc` with `objc2` in other projects


madsmtm:
As part of moving towards a (somewhat stable) `objc2 v0.3`, I'd like to submit PRs to other projects to set an ecosystem-wide precedence for using this crate, and to see which patterns people are actually using that `objc2` doesn't (or doesn't easily) help with.

Popular crates on `crates.io`:
- [ ] [`cocoa`/`cocoa-foundation`](https://github.com/servo/core-foundation-rs)
- [ ] [`winit`](https://github.com/rust-windowing/winit)
- [ ] [`glutin`](https://github.com/rust-windowing/glutin)
- [ ] [`clipboard-files`](https://github.com/betamos/clipboard-files)
- [ ] [`core-video-sys`](https://github.com/luozijun/rust-core-video-sys) https://github.com/LuoZijun/rust-core-video-sys/pull/23
- [ ] [`metal`](https://github.com/gfx-rs/metal-rs)
- [ ] [`gfx` backends](https://github.com/gfx-rs/gfx)
- [ ] [`wgpu-hal`](https://github.com/gfx-rs/wgpu/tree/master/wgpu-hal)
- [ ] [`arboard`](https://github.com/1Password/arboard)
- [ ] [`copypasta`](https://github.com/alacritty/copypasta)
- [ ] [`local_config`](https://github.com/rust-locale/locale_config)
- [ ] [`rfd`](https://github.com/PolyMeilex/rfd)
- [ ] [`glfw`](https://github.com/PistonDevelopers/glfw-rs)
- [ ] [`wry`](https://github.com/tauri-apps/wry)
- [ ] [`tao`](https://github.com/tauri-apps/tao)
- [ ] ~`offscreen_gl_context`~ now `suftman`.
- [ ] [`trash`](https://github.com/Byron/trash-rs)
- [ ] [`skia`](https://github.com/rust-skia/rust-skia)
- [ ] [`native-dialog`](https://github.com/balthild/native-dialog-rs)
- [ ] [`window_clipboard`](https://github.com/hecrj/window_clipboard)
- [ ] [`vulkano`](https://github.com/vulkano-rs/vulkano)
- [ ] [`miniquad`](https://github.com/not-fl3/miniquad)
- [ ] [`surfman`](https://github.com/servo/surfman)
- [ ] [`display-link`](https://github.com/BrainiumLLC/display-link)
[...]

PR #305, opened 2022-12-17
Remove libstd from objc-sys and block-sys

notgull:
This removes `libstd` as a dependency for `objc-sys` and `block-sys`, by replacing `std::os::raw` with `core::ffi`, which contains the necessary types.

This bumps to MSRV to at least 1.64, but it looks like this package doesn't have an MSRV policy yet (#203), so I hope that's not an issue.

### DIFF
--- a/crates/block-sys/src/lib.rs
+++ b/crates/block-sys/src/lib.rs
@@ -24,6 +24,9 @@
 #![cfg_attr(feature = "unstable-docsrs", feature(doc_auto_cfg, doc_cfg_hide))]
 #![cfg_attr(feature = "unstable-docsrs", doc(cfg_hide(doc)))]
 
+#[cfg(feature = "std")]
+extern crate std;
+
 // Ensure linkage actually happens
 #[cfg(feature = "gnustep-1-7")]
 extern crate objc_sys as _;
--- a/crates/block-sys/src/lib.rs
+++ b/crates/block-sys/src/lib.rs
@@ -409,7 +409,7 @@ pub struct Block_byref_extended {
     pub layout: *const c_char,
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "std"))]
 mod tests {
     use super::*;
     use core::ptr;
--- a/crates/block-sys/src/lib.rs
+++ b/crates/block-sys/src/lib.rs
@@ -24,11 +24,6 @@
 #![cfg_attr(feature = "unstable-docsrs", feature(doc_auto_cfg, doc_cfg_hide))]
 #![cfg_attr(feature = "unstable-docsrs", doc(cfg_hide(doc)))]
 
-extern crate std;
-
-#[cfg(not(feature = "std"))]
-compile_error!("The `std` feature currently must be enabled.");
-
 // Ensure linkage actually happens
 #[cfg(feature = "gnustep-1-7")]
 extern crate objc_sys as _;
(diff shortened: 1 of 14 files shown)


madsmtm:
Yeah, this crate doesn't have a policy yet so this wouldn't be a breaking change; however, I would still like it to work with `winit`, which is currently at `1.60.0` with a rough policy of "every Rust version released in the last 6 months".

notgull:
We could use `libc` or `cty` to circumvent the MSRV bump, if you're fine with the extra dependency.

madsmtm:
I think I've also sort-of tried to keep the amount of dependencies down.

For now, I've kinda deemed that the downsides of "bump Rust version" or "added dependency" outweigh the benefit "`no_std` support", since these are basically only useful on Apple platforms, where the standard library is already present.

But in a few months, when 1.64 is in `winit`'s MSRV, I'll probably do this change.

silvanshade:
> But in a few months, when 1.64 is in `winit`'s MSRV, I'll probably do this change.

One thing to note is that the Tauri project seems to want to keep an [MSRV of 1.61](https://github.com/tauri-apps/tao/pull/655#issuecomment-1364752455). I'm not sure if I would let that be a deciding factor, just something to be aware of considering #174.

notgull:
Whoops, I was deleting forks and I accidentally deleted the fork with this change.

Now that `winit` has an MSRV of 1.64, would this change be accepted now?

crates/block-sys/src/lib.rs
```diff
@@ -24,11 +24,9 @@
 #![cfg_attr(feature = "unstable-docsrs", feature(doc_auto_cfg, doc_cfg_hide))]
 #![cfg_attr(feature = "unstable-docsrs", doc(cfg_hide(doc)))]
 
+#[cfg(feature = "std")]
 extern crate std;
 
-#[cfg(not(feature = "std"))]
-compile_error!("The `std` feature currently must be enabled.");
-
```
madsmtm, line 31: Please do roughly the same setup as I did in [this PR to the `url` project](https://github.com/servo/rust-url/pull/722/files#diff-8d002470387f8d121e1b1c8883fe8c7c79b8309ad368499e03af0a6d1358abb5) (for both `alloc` and `std`, since neither are required).

madsmtm (changes requested):
Yeah, I'd be fine with a MSRV bump, mostly because some things are much easier to express with GATs (from Rust 1.65).

Another thing that would be nice is [more capabilities for `const fn`](https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html#more-capabilities-for-const-fn), esp. the `fn` pointer handling for eventually allowing safe method overriding in `declare_class!`. I'd also really like to start using `ptr::cast_mut`!

---

All in all we don't really gain _that_ much from the bump yet though, since `objc2` still needs `std::sync::Once` for `declare_class!`.
I think perhaps we could avoid that completely with a helper like `objc2::__macro_helpers::CachedClass` + (ab-)using the fact that the `objc` runtime is locked while creating classes? Or maybe we'd need to [re-implement parts of `std::sync::Once`](https://github.com/rust-lang/rust/blob/da636956dd13638a5582f9dcd4699559d85ac2d5/library/std/src/sys_common/once/queue.rs)? Or perhaps do so using functionality from `libdispatch`?

Do you know what the recommended approach is for `std::sync::Once`-like functionality in `no_std`?

notgull:
Not as enthusiastic about this anymore.

madsmtm:
No problem, I'll probably get around to doing it myself at some point, then I'll use the code here

PR #326, opened 2023-01-09
icrate method skips for ASAuthorization*

silvanshade:
I was trying to use the header translator on my system and needed to add these entries to the `translation-config.toml`.

I am running on macOS 13.0.1 with Xcode 14.2. When I run the header translator there are several changes to the generated output, which I'm guessing may be due to running the translator with more recent SDKs.

### DIFF
--- a/crates/header-translator/translation-config.toml
+++ b/crates/header-translator/translation-config.toml
@@ -743,3 +743,12 @@ skipped = true
 [class.CAMediaTimingFunction.methods]
 functionWithControlPoints = { skipped = true }
 initWithControlPoints = { skipped = true }
+
+[class.ASAuthorizationProviderExtensionLoginConfiguration.methods]
+setCustomAssertionRequestBodyClaims_returningError = { skipped = true }
+setCustomAssertionRequestHeaderClaims_returningError = { skipped = true }
+setCustomLoginRequestBodyClaims_returningError = { skipped = true }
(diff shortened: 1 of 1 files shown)


madsmtm:
Yeah, that's probably it.

I'm currently using the XCode `14.0.1` SDK, which is also [GitHub actions' current default](https://github.com/actions/runner-images/blob/e6fcf60b8e6c0f80a065327eaefe836881c28b68/images/macos/macos-12-Readme.md#xcode) (is there an option to change that?).

If you want, you could [download](https://developer.apple.com/download/all/?q=xcode) that release of XCode, and then point `header-translator` to that path instead (that's why it takes that as it's first parameter).

Though I could also just update the XCode version if it's a bother.

silvanshade:
> I'm currently using the XCode `14.0.1` SDK, which is also [GitHub actions' current default](https://github.com/actions/runner-images/blob/e6fcf60b8e6c0f80a065327eaefe836881c28b68/images/macos/macos-12-Readme.md#xcode) (is there an option to change that?).

It looks like you can set which version is used using one of these two methods:


```bash
# xcode-select
sudo xcode-select -s /Applications/Xcode_14.2.app/Contents/Developer
```

or

```bash
# setting the DEVELOPER_DIR env var
env:
    DEVELOPER_DIR: /Applications/Xcode_14.2.app/Contents/Developer
```

PR #327, opened 2023-01-09
icrate method skips for WebKit

silvanshade:
I made an initial attempt at autogenerating `WebKit` using the header-translator and this PR includes the modifications to the `translation-config.toml` I needed. I haven't tried to use the generated output yet so I haven't included that in this PR.

I'm interesting in using this for a tauri project I'm working on that requires interfacing with the platform-specific webview (which, in the case of macos, involves `WKWebView`.

More generally, I'm interested in #174, specifically for `wry`, `tao`, and `tauri`, but potentially other projects as well if you're still looking for help with that.


crates/header-translator/translation-config.toml
```diff
@@ -739,7 +746,102 @@ skipped = true
...
+getAvailableContentRuleListIdentifiers = { skipped = true }
+lookUpContentRuleListForIdentifier_completionHandler = { skipped = true }
+removeContentRuleListForIdentifier_completionHandler = { skipped = true }
+
+[class.WebHistory.methods]
+loadFromURL_error = { skipped = true }
+saveToURL_error = { skipped = true }
+
+[protocol.WebPlugInViewFactory.methods]
+plugInViewWithArguments = { skipped = true }
```
madsmtm, line 843: Could you document on each of these what the reason for their exclusion is? Since that will make it easier to include them in the future, once the issue is resolved.

madsmtm (changes requested):
Thanks, I'm definitely interested in supporting this framework as well!

The current setup is a bit bespoke, so feel free to point out and fix things that are annoying about the tool!

> I haven't tried to use the generated output yet so I haven't included that in this PR.

I think I'd prefer to have that before I merge this PR (will require an accompanying PR in [`icrate-generated`](https://github.com/madsmtm/icrate-generated), which is, yeah, annoying).

> potentially other projects as well if you're still looking for help with that.

Definitely, though probably not quite yet, I think it's a little too early for `icrate` for us to do that ;) (it's a difficult balance between "when is `icrate` ready" and "when will it cause users the least pain when needing to upgrade").

crates/header-translator/translation-config.toml
silvanshade: Sure, I'll add a comment for those.

I should have actually marked the PR as a draft until I can verify that the output actually works but I'll go back and double check all of these and note once it's ready.

silvanshade:
@madsmtm It seems a lot of the remaining method skip directives I needed to add were due to the compiler giving errors due to `MessageReceiver` not being implemented for `Option<Allocated<T>>`. I was looking at adding this instance and was wondering if this is the correct way to do it:

```rust
impl<T: Message + ?Sized> Sealed for Option<Allocated<T>> {}

unsafe impl<T: Message + ?Sized> MessageReceiver for Option<Allocated<T>> {
    type __Inner = T;

    #[inline]
    fn __as_raw_receiver(self) -> *mut Object {
        Allocated::option_into_ptr(self).cast()
    }
}
```

If so, I'll go ahead and add this to the PR and update the config.

madsmtm:
Could you give an example of a generated method that causes the error?

silvanshade:
> Could you give an example of a generated method that causes the error?

Sure. Actually looking at the generated output just now there aren't as many as I thought, and they are all related to init methods for various DOM event classes. When I was trying to generate the bindings, it looked like I was getting errors on all of the methods for those classes (which I elided below). But I think just skipping those init methods would be enough. I also note that most of these classes and methods are listed as deprecated too, but since we aren't really excluding those in principle at this point, might as well try to fix it for completeness sake.

Note that this is the generated output after I added those impls (I also had to add up to 16 argument impls for `EncodeArguments` and `MessageArguments`).

```rust
extern_methods!(
    unsafe impl DOMEvent {
        #[method(initEvent:canBubbleArg:cancelableArg:)]
        pub unsafe fn initEvent_canBubbleArg_cancelableArg(
            this: Option<Allocated<Self>>,
            eventTypeArg: Option<&NSString>,
            canBubbleArg: bool,
            cancelableArg: bool,
        );
    }
);
```

```rust
extern_methods!(
    unsafe impl DOMKeyboardEvent {
        #[method(initKeyboardEvent:canBubble:cancelable:view:keyIdentifier:location:ctrlKey:altKey:shiftKey:metaKey:altGraphKey:)]
        pub unsafe fn initKeyboardEvent_canBubble_cancelable_view_keyIdentifier_location_ctrlKey_altKey_shiftKey_metaKey_altGraphKey(
            this: Option<Allocated<Self>>,
            type_: Option<&NSString>,
            canBubble: bool,
            cancelable: bool,
            view: Option<&DOMAbstractView>,
            keyIdentifier: Option<&NSString>,
            location: c_uint,
            ctrlKey: bool,
            altKey: bool,
            shiftKey: bool,
            metaKey: bool,
            altGraphKey: bool,
        );

        #[method(initKeyboardEvent:canBubble:cancelable:view:keyIdentifier:location:ctrlKey:altKey:shiftKey:metaKey:)]
        pub unsafe fn initKeyboardEvent_canBubble_cancelable_view_keyIdentifier_location_ctrlKey_altKey_shiftKey_metaKey(
            this: Option<Allocated<Self>>,
            type_: Option<&NSString>,
            canBubble: bool,
            cancelable: bool,
            view: Option<&DOMAbstractView>,
            keyIdentifier: Option<&NSString>,
            location: c_uint,
            ctrlKey: bool,
            altKey: bool,
            shiftKey: bool,
            metaKey: bool,
        );

        #[method(initKeyboardEvent:canBubble:cancelable:view:keyIdentifier:keyLocation:ctrlKey:altKey:shiftKey:metaKey:altGraphKey:)]
        pub unsafe fn initKeyboardEvent_canBubble_cancelable_view_keyIdentifier_keyLocation_ctrlKey_altKey_shiftKey_metaKey_altGraphKey(
            this: Option<Allocated<Self>>,
            type_: Option<&NSString>,
            canBubble: bool,
            cancelable: bool,
            view: Option<&DOMAbstractView>,
            keyIdentifier: Option<&NSString>,
            keyLocation: c_uint,
            ctrlKey: bool,
            altKey: bool,
            shiftKey: bool,
            metaKey: bool,
            altGraphKey: bool,
        );

        #[method(initKeyboardEvent:canBubble:cancelable:view:keyIdentifier:keyLocation:ctrlKey:altKey:shiftKey:metaKey:)]
        pub unsafe fn initKeyboardEvent_canBubble_cancelable_view_keyIdentifier_keyLocation_ctrlKey_altKey_shiftKey_metaKey(
            this: Option<Allocated<Self>>,
            type_: Option<&NSString>,
            canBubble: bool,
            cancelable: bool,
            view: Option<&DOMAbstractView>,
            keyIdentifier: Option<&NSString>,
            keyLocation: c_uint,
            ctrlKey: bool,
            altKey: bool,
            shiftKey: bool,
            metaKey: bool,
        );
    }
);
```

```rust
extern_methods!(
    unsafe impl DOMMouseEvent {
        #[method(initMouseEvent:canBubble:cancelable:view:detail:screenX:screenY:clientX:clientY:ctrlKey:altKey:shiftKey:metaKey:button:relatedTarget:)]
        pub unsafe fn initMouseEvent_canBubble_cancelable_view_detail_screenX_screenY_clientX_clientY_ctrlKey_altKey_shiftKey_metaKey_button_relatedTarget(
            this: Option<Allocated<Self>>,
            type_: Option<&NSString>,
            canBubble: bool,
            cancelable: bool,
            view: Option<&DOMAbstractView>,
            detail: c_int,
            screenX: c_int,
            screenY: c_int,
            clientX: c_int,
            clientY: c_int,
            ctrlKey: bool,
            altKey: bool,
            shiftKey: bool,
            metaKey: bool,
            button: c_ushort,
            relatedTarget: Option<&DOMEventTarget>,
        );
    }
);
```

```rust
extern_methods!(
    unsafe impl DOMMutationEvent {
        #[method(initMutationEvent:canBubble:cancelable:relatedNode:prevValue:newValue:attrName:attrChange:)]
        pub unsafe fn initMutationEvent_canBubble_cancelable_relatedNode_prevValue_newValue_attrName_attrChange(
            this: Option<Allocated<Self>>,
            type_: Option<&NSString>,
            canBubble: bool,
            cancelable: bool,
            relatedNode: Option<&DOMNode>,
            prevValue: Option<&NSString>,
            newValue: Option<&NSString>,
            attrName: Option<&NSString>,
            attrChange: c_ushort,
        );
    }
);

```

```rust
extern_methods!(
    unsafe impl DOMOverflowEvent {
        #[method(initOverflowEvent:horizontalOverflow:verticalOverflow:)]
        pub unsafe fn initOverflowEvent_horizontalOverflow_verticalOverflow(
            this: Option<Allocated<Self>>,
            orient: c_ushort,
            horizontalOverflow: bool,
            verticalOverflow: bool,
        );
    }
);
```

```rust
extern_methods!(
    unsafe impl DOMUIEvent {
        #[method(initUIEvent:canBubble:cancelable:view:detail:)]
        pub unsafe fn initUIEvent_canBubble_cancelable_view_detail(
            this: Option<Allocated<Self>>,
            type_: Option<&NSString>,
            canBubble: bool,
            cancelable: bool,
            view: Option<&DOMAbstractView>,
            detail: c_int,
        );
    }
);
```

```rust
extern_methods!(
    unsafe impl DOMWheelEvent {
        #[method(initWheelEvent:wheelDeltaY:view:screenX:screenY:clientX:clientY:ctrlKey:altKey:shiftKey:metaKey:)]
        pub unsafe fn initWheelEvent_wheelDeltaY_view_screenX_screenY_clientX_clientY_ctrlKey_altKey_shiftKey_metaKey(
            this: Option<Allocated<Self>>,
            wheelDeltaX: c_int,
            wheelDeltaY: c_int,
            view: Option<&DOMAbstractView>,
            screenX: c_int,
            screenY: c_int,
            clientX: c_int,
            clientY: c_int,
            ctrlKey: bool,
            altKey: bool,
            shiftKey: bool,
            metaKey: bool,
        );
    }
);
```

madsmtm:
Ah, okay.

So this is a bit gnarly, but the gist is that some methods like `init`, `alloc`, `new`, `newWithParam:`, and so on belong to a "family", which describe the calling convention of said method.

It just so happens that this property, the "family", and therefore the calling convention, is always* possible to know just by looking at the selector. Which is what I use in `msg_send_id!` and what `header-translator` uses to determine when to use `#[method_id(...)]` vs `#[method(...)]`.

Notice the asterisk above? The thing is, the `init`-family has a special case for non-pointer return values, in which case it falls back to just being a regular method.

In your case, the methods return `void`, but `header-translator` still thinks it's dealing with an `init` method, so it happily emits the `this: Option<Allocated<Self>>`. This is wrong, since that will consume the item, making it impossible to use afterwards, and would also causing a leak - which is exactly why the type-system later catches it.

I'll have a look at fixing it, essentially we should just output `#[method(...)]` and `&self` instead, give me a day or two (and just keep them disabled for now).

See [method families](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#method-families) and [semantics of init](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#semantics-of-init) in clang's documentation of ARC for more details.

> When I was trying to generate the bindings, it looked like I was getting errors on all of the methods for those classes

Yeah, that's an issue with `extern_methods!` that's _possible_ to fix, but has a certain probability of degrading other diagnostics. I might try to do it at some point though.

silvanshade:
By the way, I was able to update this some and add support for `JavaScriptCore`.

There are some things I needed to skip from that framework (mentioned in the config) having to do with errors about `bool` not supporting encode. I read the docs about `Encode` and saw handling about `bool` and `BOOL` but wasn't really sure how to resolve the issue. (Also trying to compile with those definitions changed manually to `BOOL` gave some warnings about the struct not being FFI safe IIRC, so that's another potential issue).

Adding `Security` still seems problematic since the header-translator doesn't even finish processing and gives several errors along the way, so will have to revisit that later.

I also note that the PR is failing on the "compiler UI" tests [here](https://github.com/madsmtm/objc2/actions/runs/3915093048/jobs/6692924044) but I'm not sure what to do about that.

madsmtm:
> Adding `Security` still seems problematic since the header-translator doesn't even finish processing and gives several errors along the way, so will have to revisit that later.

I think I would actually like to not include support for `JavaScriptCore` just quite yet, if that's feasible? Mostly because that includes a large [pure-C API](https://developer.apple.com/documentation/javascriptcore/c_javascriptcore_api), much in the same way as `CoreFoundation`, which `header-translator` is not yet really equipped to handle.

> By the way, I was able to update this some and add support for `JavaScriptCore`.
>
> There are some things I needed to skip from that framework (mentioned in the config) having to do with errors about `bool` not supporting encode. I read the docs about `Encode` and saw handling about `bool` and `BOOL` but wasn't really sure how to resolve the issue. (Also trying to compile with those definitions changed manually to `BOOL` gave some warnings about the struct not being FFI safe IIRC, so that's another potential issue).

Yeah, I can understand why you're confused; the problem is that the function typedefs used in those structs (`JSObjectHasPropertyCallback`, `JSObjectSetPropertyCallback`, `JSObjectDeletePropertyCallback` and `JSObjectHasInstanceCallback`) return C99's `bool` instead of Objective-C's `BOOL`, which, again, outside what `header-translator` is really set up to handle just yet. So special-casing it is fine for now.

> I also note that the PR is failing on the "compiler UI" tests [here](https://github.com/madsmtm/objc2/actions/runs/3915093048/jobs/6692924044) but I'm not sure what to do about that.

I suspect in this case it's caused by a nightly compiler update which changes diagnostics. If you run:
```
TRYBUILD=overwrite cargo +nightly run --features=run --bin test-ui
```

The test cases will be automatically updated. If you want, you can make a separate PR for updating this, otherwise I'll just do it myself soon-ish.

silvanshade:
> I think I would actually like to not include support for `JavaScriptCore` just quite yet, if that's feasible? Mostly because that includes a large [pure-C API](https://developer.apple.com/documentation/javascriptcore/c_javascriptcore_api), much in the same way as `CoreFoundation`, which `header-translator` is not yet really equipped to handle.

Yeah that's fine. I think some parts of Tauri may need access to the javascript context stuff which was the main motivation for including it (aside for completeness) but I think that usage is limited enough that we could just deal with it through some separate manual definitions for now. I'll update the PR and remove that for now.

crates/icrate/src/lib.rs
```diff
@@ -42,5 +42,9 @@ pub mod CoreAnimation;
 pub mod CoreData;
 #[cfg(feature = "Foundation")]
 pub mod Foundation;
+#[cfg(feature = "JavaScriptCore")]
+pub mod JavaScriptCore;
 #[cfg(feature = "Metal")]
 pub mod Metal;
+#[cfg(feature = "WebKit")]
+pub mod WebKit;
```
madsmtm, line 50: Also add an entry in `header-translator/src/data/...`, to ease the process of making methods in `WebKit` safe

madsmtm:
Do you want to try rebasing this PR now that the linked issues have been merged?

silvanshade:
> Do you want to try rebasing this PR now that the linked issues have been merged?
> 
> (And just keep this and the generated output in one PR, again, sorry for the confusion earlier)

Okay, just rebased.

crates/header-translator/translation-config.toml
```diff
@@ -827,3 +830,24 @@ skipped = true
...
+globalContext = { skipped = true }
+
+# error: translator assertion failure: invalid error nullability (left: Unspecified) (Right: Nullable)
+[class.WebHistory.methods]
+loadFromURL_error = { skipped = true }
+saveToURL_error = { skipped = true }
+
+[protocol.WebPlugInViewFactory.methods]
+# error: class methods are not supported in `extern_protocol!`
+plugInViewWithArguments = { skipped = true }
```
madsmtm, line 845: Nit: Move this next to the ones in https://github.com/madsmtm/objc2/blob/2bd99a848d6ea31d279b8dd88e3e1e6a1467b58a/crates/header-translator/translation-config.toml#L321-L339

crates/header-translator/translation-config.toml
```diff
@@ -827,3 +830,24 @@ skipped = true
...
+
+[class.WebScriptObject.methods]
+# Needs `JSObjectRef` from JavaScriptCore
+JSObject = { skipped = true }
```
madsmtm, line 849: Nit: Put this section next to `WebFrame.methods`

crates/objc2-encode/src/encode.rs
```diff
@@ -712,6 +712,10 @@ encode_args_impl!(A, B, C, D, E, F, G, H, I);
 encode_args_impl!(A, B, C, D, E, F, G, H, I, J);
 encode_args_impl!(A, B, C, D, E, F, G, H, I, J, K);
 encode_args_impl!(A, B, C, D, E, F, G, H, I, J, K, L);
+encode_args_impl!(A, B, C, D, E, F, G, H, I, J, K, L, M);
+encode_args_impl!(A, B, C, D, E, F, G, H, I, J, K, L, M, N);
+encode_args_impl!(A, B, C, D, E, F, G, H, I, J, K, L, M, N, O);
+encode_args_impl!(A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P);
```
madsmtm, line 718: Add changelog entry for this, and update the docs that say it is implemented for 12 arguments

crates/objc2/src/message/mod.rs
```diff
@@ -565,6 +565,72 @@ message_args_impl!(
...
+    h: H,
+    i: I,
+    j: J,
+    k: K,
+    l: L,
+    m: M,
+    n: N,
+    o: O,
+    p: P
+);
```
madsmtm, line 633: Add changelog entry for this, and update the docs that say it is implemented for 12 arguments

crates/objc2/src/message/mod.rs
```diff
@@ -565,6 +565,72 @@ message_args_impl!(
     k: K,
     l: L
 );
+message_args_impl!(
```
madsmtm, line 568: Also update `method_decl_impl` in `declare.rs`

crates/icrate/src/WebKit/mod.rs
```diff
@@ -0,0 +1,7 @@
+#[path = "../generated/WebKit/mod.rs"]
+mod generated;
+
+pub use self::generated::*;
+
+#[cfg_attr(feature = "apple", link(name = "WebKit", kind = "framework"))]
```
madsmtm, line 6: Nit: Always link to `WebKit`, similar to e.g. `AuthenticationServices`
```suggestion
#[link(name = "WebKit", kind = "framework")]
```

madsmtm (changes requested):
Thanks, have a few concerns regarding mostly the added support for 16-argument selectors.

madsmtm:
Replaced by https://github.com/madsmtm/objc2/pull/355

Issue #311, opened 2022-12-23
Improve compilation time of `icrate` 


madsmtm:
Probably by making everything hidden behind selective cargo features like `Foundation_NSMutableArray` (which would transitively enable `Foundation_NSArray` and `Foundation_NSObject`).

madsmtm:
So far I've determined that the compilation time has increased quite a lot since [I last looked at it](https://github.com/madsmtm/objc2/pull/264#issuecomment-1299555869) - it's now at ~48 seconds.

Further investigation shows that the `extern_methods!` declaration is the most time-consuming (accounting for ~80%) (not so very surprising). But even if I change things so that we manually output `MessageReceiver::send_message` (and equivalent for `msg_send_id!`), the time only drops by a small amount (~45s), so the increased time comes from https://github.com/madsmtm/objc2/pull/198 maybe?

madsmtm:
Indeed, doing `[profile.dev] debug-assertions = false` immediately cuts the time down to ~33s

madsmtm:
Noting down `-Ztime-passes` and [`-Zself-profile`](https://doc.rust-lang.org/unstable-book/compiler-flags/self-profile.html) for my own future reference

silvanshade:
> So far I've determined that the compilation time has increased quite a lot since [I last looked at it](https://github.com/madsmtm/objc2/pull/264#issuecomment-1299555869) - it's now at ~48 seconds.

I've been working on #327, trying to get to a point where the `WebKit` bindings are usable, and so far this seems to be one of the bigger issues I'm facing. After adding the method skip directives from that PR to the translation config, compilation doesn't immediately error anymore (more on that though) but compilation seems to take a very long time. I waited several minutes a few different times and never saw it finish.

I've tried manually including only the definitions needed from the generated output, and I'm seeing some more errors now (which I'm fixing), so maybe somehow compilation times will get better if I track down the remaining issues.

However, I'm thinking the approach you mention, i.e. feature `Foundation_NSMutableArray` transitively enables features `Foundation_NSArray` and `Foundation_NSObject`, may not work out much better.

For example, just trying to include enough of the generated output to import `WKWebView` has me having to go through and import so much other unused stuff that it's probably going to end up pulling in almost all of the `WebKit` module.

[...]

PR #328, opened 2023-01-09
Put methods behind feature flags

madsmtm:
Fixes https://github.com/madsmtm/objc2/issues/311.

A few manual fixes are still necessary to handle structs and typedefs that use class types, and the user has to specify e.g. `["Foundation", "Foundation_NSString"]` instead of just `["Foundation_NSString"]`, but it should be workable.


silvanshade:
Awesome, thanks for starting on this!

silvanshade:
I made some additions in a local branch at https://github.com/silvanshade/objc2/tree/icrate-behind-feature-flags. It's able to automatically emit the features to the `icrate/Cargo.toml` (see the comment at the end of the file) although it's apparently missing some cases, e.g.:

```
feature `AppKit_NSScrubberItemView` includes `AppKit_NSScrubberArrangedView` which is neither a dependency nor another feature
```

In any case, may not be the best way to do things but feel free to use any of that if it's useful.

madsmtm:
Thanks, I started from what you had and have gotten feature emission to work now (not currently updated in-branch).

crates/header-translator/src/main.rs
```diff
@@ -129,14 +132,51 @@ fn main() {
     cache.update(&mut final_result);
     drop(span);
 
-    for (library_name, files) in final_result.libraries {
+    for (library_name, files) in &final_result.libraries {
         let _span = info_span!("writing", library_name).entered();
-        let output_path = crate_src.join("generated").join(&library_name);
+        let output_path = crate_src.join("generated").join(library_name);
```
silvanshade, line 137: Can we add this line so that the generator doesn't error if the library dir hasn't been manually created yet?

```rust
std::fs::create_dir_all(&output_path)?;
```

crates/header-translator/src/stmt.rs
```diff
@@ -879,32 +882,30 @@ impl fmt::Display for Stmt {
...
                 // TODO: Use ty.get_objc_protocol_declarations()
 
-                let macro_name = if ty.generics.is_empty() {
+                let macro_name = if generics.is_empty() {
                     "extern_class"
                 } else {
                     "__inner_extern_class"
                 };
 
                 writeln!(f, "{macro_name}!(")?;
```
silvanshade, line 891: Do we need to emit the feature attributes here too (and at line 922 below)?

```rust
if let Some(feature) = id.feature() {
    writeln!(f, "    #[cfg(feature = \"{feature}\")]")?;
}
```

This way we generate output like this:

```rust
extern_class!(
    #[cfg(feature = "WebKit_DOMAbstractView")]
    #[derive(Debug, PartialEq, Eq, Hash)]
    pub struct DOMAbstractView;

    #[cfg(feature = "WebKit_DOMAbstractView")]
    unsafe impl ClassType for DOMAbstractView {
        #[inherits(WebScriptObject, NSObject)]
        type Super = DOMObject;
    }
);
```

madsmtm: True, this is now done

crates/header-translator/src/main.rs
madsmtm: Huh, I thought I included that from your changes before.

I've included it now, at least

madsmtm:
So wrt. compilation-time improvements (approximate numbers):

| Label | `--features=Foundation` currently | `--features=Foundation` now | `--features=Foundation_all` now |
| --- | --- | --- | --- |
| `debug` check | ~21s | ~4s | ~20s |
| `debug` build | ~31s | ~5s | ~33s |
| `release` build | ~30s | ~4s | ~26s |

Tested with:
```sh
CMD=cargo +nightly build -picrate --features Foundation # or `check` or `build --release`
$CMD
cargo clean -picrate
$CMD
```

Which tells me that this is a _definite_ improvement (and not even a regression when enabling many features), and it is definitely good enough for me to actually start using it in `winit`!

PR #332, opened 2023-01-10
Various header translator refactors

madsmtm:
I basically redid the whole "parse lifetime from a string"-part, and tweaked a lot of other small things (a lot of these were small irritating things that were preventing me from doing "real" work like https://github.com/madsmtm/objc2/pull/328)


madsmtm:
So... It _felt_ like each change I made here was a step forwards, but honestly a lot of things just got more complex...

Oh well, live an learn

PR #333, opened 2023-01-10
Enable header-translator/run feature for vscode

silvanshade:
I don't know if you want to merge this but it makes developing the header-translator crate a little easier with vscode.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "rust-analyzer.cargo.features": [
+        "header-translator/run"
+    ]
+}


madsmtm (changes requested):
Hmm, I think I'd rather not, mostly because it's user-specific whether they want to download all the dependencies of `header-translator`, or whether they'd just rather like to work on e.g. `objc2`

silvanshade:
Fair enough. Just out of curiosity, how do you develop the crate locally? Maybe there is a better way to do this but it's sort of unfortunate that none of the rust-analyzer features will activate when browsing the crate unless the feature is enabled.

madsmtm:
My dev setup is terrible atm., the `rust-analyzer` extensions for my editor (Sublime Text) keeps breaking on me so I've just sorta given up. But really something I should look into fixing!

PR #335, opened 2023-01-12
Emit arguments in snake_case

silvanshade:
This PR makes the header translator output method and function arguments in snake_case.

The rationale for this change is that having the arguments in snake case works a little better overall with several rust analyzer features and idiomatic rust code.

For example, there is one rust analyzer feature that shows inlay hints for the names of function and method arguments, but it elides those hints in the case where the hint name and the actual argument name are the same. I show an example of this below. (See the method call `data_store.removeDataOfTypes_modifiedSince_completionHandler` at the bottom).

(snake_case arguments, where the hints are shown, redundantly)
snake_case

(camelCase arguments, where the hints are elided, but then we have warnings (yellow underlines) about non-snake_case identifiers)
camelCase

Similarly, there is another feature of rust-analyzer that can autocomplete method calls, filling in arguments from those already in scope, if the names and types match what the function or method expects:

```
rust-analyzer.completion.callable.snippets (default: "fill_arguments")
    Whether to add parenthesis and argument snippets when completing function.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,6 +119,7 @@ version = "0.1.0"
 dependencies = [
  "apple-sdk",
  "clang",
+ "heck",
  "proc-macro2",
  "serde",
  "toml",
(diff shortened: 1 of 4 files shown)


madsmtm (approved):
That's a great idea!

It also looks nicer in `rustdoc`.

silvanshade:
Closing in favor of #342  where I folded these changes into the last commit.

silvanshade:
Reopening so that it's easier to see changes in generated output specific to this change.

Issue #330, opened 2023-01-09
Better tooling around `header-translator` data files


madsmtm:
In https://github.com/madsmtm/objc2/pull/329 I authored a pretty large file, which naturally lead me to thinking about how that process could be improved upon:

- [ ] Some `rustdoc` fork that allows one to go through, click a button, and have the source-code updated to mark it as safe (perhaps with a comment).
- [ ] Are property accessors always safe? I know that setters are not, since they may need to maintain an invariant, but it seems like if you have a reference to a type, you should always be able to access any data it holds? Maybe just for certain types (like, we could do `NSData: Shared { ... }`, and then all `NSData` properties would become safe).
- [ ] A way to make icrate generate `#[method(myMethod)] pub(crate) fn myMethod_unchecked(...)` for instances where one would like to modify the method but keep the name the same (e.g. `NSData::bytes`)
- [ ] Checks in `header-translator` that verify that the data entry is actually used (and e.g. not declared `class X { ... }` on a protocol).

Issue #284, opened 2022-11-04
Figure out naming scheme in `icrate`


madsmtm:
`icrate` (or whatever it'll end up being named) is nearing an initial version, see https://github.com/madsmtm/objc2/pull/264.

After that, we need to figure out how we should handle name translation for enums, method names, and so on. There are a few options:
1. Use Objective-C naming scheme:
  - Class: `struct NSObject`
  - Protocol: `struct NSCopying`
  - Category: `#[doc = "NSAccessibility"] impl NSObject`
  - Method: `fn someMethod(arg: i32)`, `fn someMethod_withExtraParam(arg: i32, extra: u32)`
  - Method returning error: `fn aMethod_error(arg: i32) -> Result<(), Id<NSError>>`
  - Method with callback: `async fn aMethodWithCompletionHandler(f: &Block<(), ()>)`, see also #279
  - Enum: `struct NSAccessibilityAnnotationPosition(NSInteger)`
  - Enum constant: `const NSAccessibilityAnnotationPositionFullRange: NSAccessibilityAnnotationPosition = ...`
  - Struct: `struct NSDirectionalEdgeInsets { fields* }`
  - Constant: `const NSModalResponseStop: NSModalResponse = 1000`
  - Typedef: `struct NSBitmapImageRepPropertyKey(NSString)`
  - Static: `extern "C" { static NSImageCurrentFrameDuration: NSBitmapImageRepPropertyKey }`
  - Function: `extern "C" { fn NSBitsPerPixelFromDepth(depth: NSWindowDepth) -> NSInteger }`
2. Use a scheme [similar to Swift's](https://github.com/apple/swift-evolution/blob/main/proposals/0005-objective-c-name-translation.md)
  - Class: Same, sometimes renamed?
  - Protocol: Same
  - Category: Same
  - Method: `fn some(method: i32)`, `fn some_with(method: i32, extraParam: u32)` (name is significantly stripped)
  - Method returning error: `fn a(arg: i32) -> Result<(), Id<NSError>>` (`error` postfix is removed)
  - Method with callback: `async fn a()` (`WithCompletionHandler` stripped)
  - Enum: `struct NSAccessibilityAnnotationPosition(NSInteger)`, Swift usually creates a nested items like `NSAccessibility::AnnotationPosition`, but that is not possible in Rust
  - Enum constant: `impl NSAccessibilityAnnotationPosition { const fullRange: Self = ... }`
[...]

PR #336, opened 2023-01-12
Allow multiple colons in selectors

madsmtm:
Should help cutting down on the amount of work https://github.com/madsmtm/objc2/pull/327 and futures has to do. Tagging @silvanshade.

### DIFF
--- a/crates/header-translator/translation-config.toml
+++ b/crates/header-translator/translation-config.toml
@@ -774,11 +774,6 @@ skipped = true
 [class.NSMutableData.methods.mutableBytes]
 skipped = true
 
-# Selector contains multiple colons after each other, like `::::`
-[class.CAMediaTimingFunction.methods]
-functionWithControlPoints = { skipped = true }
-initWithControlPoints = { skipped = true }
-
 # Uses __autoreleasing in a typedef, which I'm unsure how to handle
 [typedef.MTLAutoreleasedArgument]
 skipped = true
--- a/crates/objc2-proc-macros/CHANGELOG.md
+++ b/crates/objc2-proc-macros/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased - YYYY-MM-DD
 
+### Fixed
+* Allow all types of tokens in internal macro.
+
 
 ## 0.1.0 - 2022-07-19
 
--- a/crates/objc2-proc-macros/src/lib.rs
+++ b/crates/objc2-proc-macros/src/lib.rs
@@ -24,27 +24,19 @@ use proc_macro::Literal;
 use proc_macro::TokenStream;
 use proc_macro::TokenTree;
 
-/// Quick n' dirty way to extract the idents given by the `sel!` macro.
+/// Extract all identifiers in the given tokenstream.
 fn get_idents(input: TokenStream) -> impl Iterator<Item = Ident> {
     input.into_iter().flat_map(|token| {
-        if let TokenTree::Group(group) = token {
-            group
-                .stream()
-                .into_iter()
-                .map(|token| {
-                    if let TokenTree::Ident(ident) = token {
-                        ident
-                    } else {
-                        panic!("Expected ident, got {token:?}")
-                    }
-                })
-                .collect::<Vec<_>>()
-                .into_iter()
-        } else if let TokenTree::Ident(ident) = token {
-            vec![ident].into_iter()
-        } else {
-            panic!("Expected group or ident, got {token:?}")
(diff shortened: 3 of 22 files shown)


silvanshade:
Thanks for starting on this. I was taking a look at the changes and although it seems like it should fix some of the issues with methods that currently need to be skipped there may still be some problems that remain.

When I was trying this PR locally, I still ran into problems with methods that get translated as duplicate names. For instance, we have situations like this:

```rust
extern_methods!(
    #[cfg(feature = "WebKit_WebView")]
    unsafe impl WebView {
        #[method(goForward)]
        pub unsafe fn goForward(&self) -> bool;
    }
);

extern_methods!(
    /// WebIBActions
    #[cfg(feature = "WebKit_WebView")]
    unsafe impl WebView {
        #[method(goForward:)]
        pub unsafe fn goForward(&self, sender: Option<&Object>);
    }
);
```

I also notice that in the examples that now translate in the PR, all of the trailing `:` are collapsed:

```rust
    unsafe impl CAMediaTimingFunction {
        #[method_id(@__retain_semantics Other functionWithControlPoints::::)]
        pub unsafe fn functionWithControlPoints(
            c1x: c_float,
            c1y: c_float,
            c2x: c_float,
            c2y: c_float,
        ) -> Id<Self, Shared>;
    }
```

So it seems we may run into a general issue with disambiguation, say if we have a selector like `foo:bar` and another one like `foo::bar`, and then again similarly for selectors with prefixed or suffixed colons.

I guess the problem with just always naively translating `:` as `_` in the method name is that it looks kind of ugly since so many of these methods would now end with a trailing `_`, and also the reserved keywords are translated as `k` -> `k_` so then that becomes an additional source of ambiguity.

Maybe an alternative approach that would resolve these potential issues would be to adopt a strategy like the following:

1. translate keywords using only raw identifiers (e.g., `type` -> `r#type`)
2. if there are multiple methods `m0` and `m1` that would be translated to the same method name, disambiguate them by translating all of the instances of `:` directly as `_` (including multiple subsequent `_` if necessary)

(2) should not be too difficult, but one potential downside is that it could lead to an additional source of breakage with regard to #338 if a new ambiguating method gets added in an updated SDK.

Another possibility might be to allow specifying a particular renaming for a given selector directly in the translation config file and then just manually patching up these cases. I suppose the ambiguous cases are probably rare enough that that would suffice and would maybe lead to nicer names anyway (at the potential of some slight confusion, at least until or unless the API docs are linked into the generated output).

Any thoughts on whether this might work or if there is a better way to handle these issues in general?

madsmtm:
> translate keywords using only raw identifiers

That's a great idea, we should definitely do that!

> if there are multiple methods `m0` and `m1` that would be translated to the same method name, disambiguate them by translating all of the instances of `:` directly as `_` (including multiple subsequent `_` if necessary)

We could use some kind of ambiguity detection anyhow for duplicate class & instance methods. I would guess it would be easiest as a post-processing step in `cache.rs`, if you want to try have a stab at it?

> it could lead to an additional source of breakage with regard to https://github.com/madsmtm/objc2/issues/338 if a new ambiguating method gets added in an updated SDK.

Good point, though probably quite unlikely (and if it's the case, we can fix it manually just like the other breakage there).

> allow specifying a particular renaming for a given selector

I've been considering this anyhow in https://github.com/madsmtm/objc2/issues/330, and will also be necessary when we start doing https://github.com/madsmtm/objc2/issues/284.

---

Really, I'm unsure of how common it is that method names are duplicated, which is kind of required knowledge to figure out which fix is the better solution? Perhaps one could run a test script against all frameworks in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk`, and figure it out? But then again, it might just be easier to fix the problem properly and never again worry about it.

madsmtm:
We could probably also use `proc_macro2` for translating keywords to identifiers

silvanshade:
> We could use some kind of ambiguity detection anyhow for duplicate class & instance methods. I would guess it would be easiest as a post-processing step in `cache.rs`, if you want to try have a stab at it?

Okay, I will give it a shot.

I did briefly try to modify the translation of the selectors earlier, specifically removing the trimming of trailing colons [here](https://github.com/madsmtm/objc2/blob/d6867b77ce21ab116952f626a7ac28ead1e9b1ee/crates/header-translator/src/method.rs#L168) but after doing that and running the translator again it almost immediately errored out with some problem about argument analysis IIRC.

But maybe a post processing step in `cache.rs` should avoid this?

madsmtm:
Perhaps because after the function name changed, `ClassData::get_method_data` in `stmt.rs` now fails finding the method skips that we've declared in `translation-config.toml`?

PR #342, opened 2023-01-14
Selector disambiguation

silvanshade:
Addresses the discussion starting from https://github.com/madsmtm/objc2/pull/336#issuecomment-1382053237


crates/header-translator/src/cache.rs
```diff
@@ -119,6 +119,36 @@ impl<'a> Cache<'a> {
...
+                    for this in methods.iter_mut() {
+                        let key = (id.clone(), this.fn_name.clone());
+                        if let Some(that) = names.get_mut(&key) {
+                            that.fn_name = that.selector.replace(':', "_");
+                            this.fn_name = this.selector.replace(':', "_");
+                        } else {
+                            names.insert(key, this);
+                        }
+                    }
+                }
```
madsmtm, line 147: You can avoid the duplication by matching like this:
```rust
Stmt::Methods { cls: id, methods, .. }
| Stmt::ProtocolDecl { id, methods, .. } =>
```

crates/header-translator/src/cache.rs
```diff
@@ -119,6 +119,36 @@ impl<'a> Cache<'a> {
     }
 
     fn update_file(&self, file: &mut File) {
+        // disambiguate duplicate names
```
madsmtm, line 122: Nit: Add a comment that this only works within a single file (which is fine to start with!)

crates/header-translator/src/cache.rs
```diff
@@ -119,6 +119,36 @@ impl<'a> Cache<'a> {
...
+                Stmt::Methods { cls, methods, .. } => {
+                    for this in methods.iter_mut() {
+                        let key = (cls.clone(), this.fn_name.clone());
+                        if let Some(that) = names.get_mut(&key) {
+                            that.fn_name = that.selector.replace(':', "_");
+                            this.fn_name = this.selector.replace(':', "_");
+                        } else {
+                            names.insert(key, this);
+                        }
+                    }
```
madsmtm, line 135: Nit: A bit better naming IMO?

```suggestion
                    for method in methods.iter_mut() {
                        let key = (cls.clone(), method.fn_name.clone());
                        if let Some(other) = names.get_mut(&key) {
                            other.fn_name = other.selector.replace(':', "_");
                            method.fn_name = method.selector.replace(':', "_");
                        } else {
                            names.insert(key, method);
                        }
                    }
```

madsmtm (commented):
Cool!

madsmtm:
I think I would prefer the `icrate-generated` changes to happen in separate commits, to make it easier to see what has changed.

crates/header-translator/src/cache.rs
silvanshade: Ah, good point!

crates/header-translator/src/cache.rs
silvanshade: Are there instances where methods for a class or protocol can potentially be split across files? Or are there some other scenarios where we would need to disambiguate?

silvanshade:
> Also, I think I would prefer the `icrate-generated` changes to happen in separate commits, to make it easier to see what has changed.

Sure, I can do that. The reason I included it here is because I was expecting the CI to fail otherwise with a complaint about changes in the generated output.

crates/header-translator/src/cache.rs
madsmtm: Indeed there are: Any object can have new "categories" defined on it at any time. But it likely won't happen, so again, that it only works within a single file is fine.

madsmtm:
I think I meant that I would like the commit that you made in `icrate-generated` to be split into two, not that the commit should be decoupled from this PR - sorry for the confusion!

madsmtm:
I went and rebased your branch to fix it, as well as adding support for instance vs. class methods - I hope you don't mind!

PR #343, opened 2023-01-14
Snake case arguments

silvanshade:
Update of https://github.com/madsmtm/objc2/pull/335 pending changes from #342.

### DIFF
--- a/crates/header-translator/src/method.rs
+++ b/crates/header-translator/src/method.rs
@@ -612,7 +612,8 @@ impl fmt::Display for Method {
 
         // Arguments
         for (param, arg_ty) in &self.arguments {
-            write!(f, "{}: {arg_ty},", handle_reserved(param))?;
+            let param = handle_reserved(&crate::to_snake_case(param));
+            write!(f, "{param}: {arg_ty},")?;
         }
         write!(f, ")")?;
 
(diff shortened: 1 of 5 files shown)


crates/header-translator/src/method.rs
```diff
@@ -618,7 +618,12 @@ impl fmt::Display for Method {
 
         // Arguments
         for (param, arg_ty) in &self.arguments {
-            write!(f, "{}: {arg_ty},", handle_reserved(param))?;
+            let param = if param == "_" {
+                String::from("_")
+            } else {
```
madsmtm, line 623: Huh, do you have an example of when this happens?

silvanshade: I think the only place it happens is in the argument for [`NSSetUncaughtExceptionHandler`](https://docs.rs/icrate/latest/icrate/Foundation/fn.NSSetUncaughtExceptionHandler.html). The reason I have to handle it specially is because `heck` converts the `"_"` to an empty string `""`.

madsmtm: Ah okay, let's keep it as-is for now then

madsmtm:
Could I get you to fix the conflicts, or do you want me to do it?

silvanshade:
> Could I get you to fix the conflicts, or do you want me to do it?

Sure I can do that.

madsmtm (approved):
Thank you!

PR #347, opened 2023-01-15
Update our nightly version manually

madsmtm:
Previously, I've had it updated automatically, so that I could help find regressions in the compiler output faster.

But that's not really tenable now that there's starting to be other contributors to the project - so let's revert to manually updating said version (just like we do with out dependencies, which this PR also updates).


crates/test-ui/ui/extern_protocol_class_method.stderr
```diff
@@ -39,17 +39,3 @@ error: class methods are not supported in `extern_protocol!`
   | |_^
   |
   = note: this error originates in the macro `$crate::__extern_protocol_method_out` which comes from the expansion of the macro `extern_protocol` (in Nightly builds, run with -Z macro-backtrace for more info)
-
-error: class methods are not supported in `extern_protocol!`
```
madsmtm, line 43: This should not have been removed, so is probably a bug in `trybuild v1.0.75`

PR #353, opened 2023-01-16
Rename "nul" files to fix checkouts on windows

silvanshade:
Cloning the repo on windows currently fails due to files in the tree named "nul".

More info [here](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file):

> Do not use the following reserved names for the name of a file:
>
> CON, PRN, AUX, NUL, COM0, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT0, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9. Also avoid these names followed immediately by an extension; for example, NUL.txt and NUL.tar.gz are both equivalent to NUL. For more information, see [Namespaces](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces).


madsmtm:
Oh right, forgot about that

PR #356, opened 2023-01-18
Add support for Accessibility framework

silvanshade:


madsmtm:
Sorry, I haven't really figured out the best process for the submodule stuff, but I'm pretty sure I would like the refs that end up in the `master` branch to be the same ones that every commit on `objc2`'s `master` branch references. So I think it would be easiest for me if you rebased the `icrate-generated` commits on top of each other in the order you would like me to merge your PRs.

So for example commit 1: `Add support for BackgroundAssets framework`
Commit 2: `Add support for Accessibility framework`, on top of commit 1
Commit 3: ... on top of commit 2

If that makes sense? (also, you don't need to open PRs against `icrate-generated`, I'll just pull directly from your branch)

madsmtm:
So effectively serializing each change (which is, yeah, a bit cumbersome)

silvanshade:
> So effectively serializing each change (which is, yeah, a bit cumbersome)

I think these should be adjusted in that sense now.

Issue #345, opened 2023-01-15
Investigate implementing `header-translator` on top of Swift's `ClangImporter`


madsmtm:
In `header-translator` we have to do a bunch of hacks to account for the fact that a lot of things are not exposed to the `libclang` API. In particular the [Swift attributes](https://clang.llvm.org/docs/AttributeReference.html#customizing-swift-import) are desirable, e.g. the new `NS_SWIFT_UI_ACTOR` will help us a _lot_ in knowing which things require `MainThreadMarker` before access!

So crazy idea: What if instead we forked/used [Swift's ClangImporter](https://github.com/apple/swift/tree/swift-5.7.2-RELEASE/lib/ClangImporter) and built directly upon their work instead? They already handle all the edge-cases, they have a system for importing `.apinotes` files, and so on.

Though that would probably require more fluency in C++ than I really have, as well as the hard work of actually figuring out how to interface with it in a nice way, and probably also a beefier computer to compile all that code than the one I currently have :/

silvanshade:
I've started some initial work in this direction [here](https://github.com/silvanshade/clang-importer-sys).

Right now the plan is to start building out the bindings to `ClangImporter` using a combination of [cxx](https://github.com/dtolnay/cxx) (for most of the declarations) and [rust-cpp](https://github.com/mystor/rust-cpp) (where some additional inline wrapper code is needed to support the `cxx` bindings).

I originally tried using [autocxx](https://github.com/google/autocxx) (which uses a modified `bindgen` to generate `cxx` directives) but ran into a problem with what I think was a stack overflow. Apparently there are some known issues where this can happen for particularly complex libraries.

It would have been nice if `autocxx` had worked but this more manual still seems manageable.

Note that I haven't tested this on macOS yet so you would need to modify the `build.rs` script and update the `TARGET` appropriately const for that case if you try to build it currently.

madsmtm:
Cool!

I think I was planning on just rewriting `header-translator` entirely in C++, but if we can write most of it in Rust and then use `ClangImporter` through `cxx` then I'm all ears.

silvanshade:
There's still a lot left to do before this this thing is usable but there's at least enough in place now to show that we can create a `ClangImporter` instance: see [here](https://github.com/silvanshade/framework-translator/blob/196604b0f938cbe721b73310768d0d2f4e7150aa/crates/clang-importer-sys/tests/swift/clang_importer.rs#L18).

madsmtm:
[...]

PR #357, opened 2023-01-18
Add support for BackgroundAssets framework

silvanshade:


crates/icrate/src/BackgroundAssets/fixes/mod.rs
```diff
@@ -0,0 +1,56 @@
+#![allow(clippy::bool_to_int_with_if)]
```
madsmtm, line 1: Did you mean to add this file?

madsmtm (approved):
Sure, this seems simple enough to support

crates/icrate/src/BackgroundAssets/fixes/mod.rs
silvanshade: Nope, I'll fix it and rebase.

silvanshade:
> Sure, this seems simple enough to support

Just out of curiosity, do you have a list of frameworks you would ultimately like to support or is it basically just "as much as possible"?

I'm also wondering what the plan is regarding more of the "core" frameworks that you alluded to earlier.

I made a quick attempt a little while ago to rewrite most of `tao` and `wry` using `icrate` (with the `WebKit` additions) and there's still quite a few parts of it that rely on definitions coming from the core-graphics crates, some of which overlap with the bits already included in `icrate.

madsmtm:
> Just out of curiosity, do you have a list of frameworks you would ultimately like to support or is it basically just "as much as possible"?

I think the plan is mostly just "if people need it". So well, at some point that might be everything.

> I'm also wondering what the plan is regarding more of the "core" frameworks that you alluded to earlier.

I'm still unsure! Ideally, I'd like to integrate with the rest of the ecosystem, and use crates like `core-foundation` and `core-graphics`; but on the other hand, when auto-generating stuff, _everything_ from a framework needs to be available for us to be able to use it; so I suspect it will still need a lot of manual fixes, and then it might just be better to re-implement it (which also gives us the benefit of getting availability attributes on those frameworks, and other improvements that might come to `icrate`).

There's also the question of how much we should try to make it safe, ideally something like `CGContextRef` would be translated to a class-like structure `CGContext` with methods and a destructor, just like it is in Swift.

I'm considering redoing `header-translator` the "proper" way, which is to use the existing code that Swift uses, see https://github.com/madsmtm/objc2/issues/345 - so possibly CoreFoundation-like structs will only happen after that.

silvanshade:
> I think the plan is mostly just "if people need it". So well, at some point that might be everything.

That seems reasonable enough. I guess what I'm wondering then is whether you mind if I just create PRs for the various frameworks as I'm able to run them through the translator and figure out what needs to be skipped for now. I think the more that is available in the crate, the more likely it might be that someone will actually use it, rather than asking for it to be supported. That also seems to provide more opportunity to figure out ways the tooling and translation process can be improved, as issues are encountered.

> I'm still unsure! Ideally, I'd like to integrate with the rest of the ecosystem, and use crates like `core-foundation` and `core-graphics`; but on the other hand, when auto-generating stuff, _everything_ from a framework needs to be available for us to be able to use it; so I suspect it will still need a lot of manual fixes, and then it might just be better to re-implement it (which also gives us the benefit of getting availability attributes on those frameworks, and other improvements that might come to `icrate`).

Yeah... I think integration would also be nice but I'm also not sure how practical it will be to coordinate given those observations.

> I'm considering redoing `header-translator` the "proper" way, which is to use the existing code that Swift uses, see #345 - so possibly CoreFoundation-like structs will only happen after that.

That sounds like a really cool idea actually!

madsmtm:
> That seems reasonable enough. I guess what I'm wondering then is whether you mind if I just create PRs for the various frameworks as I'm able to run them through the translator and figure out what needs to be skipped for now. I think the more that is available in the crate, the more likely it might be that someone will actually use it, rather than asking for it to be supported. That also seems to provide more opportunity to figure out ways the tooling and translation process can be improved, as issues are encountered.

Hmm, I think I'd like to get iOS up n' running first, since we'll probably need to do a lot of manual fixes there, and the less frameworks we have that we have to do it for, the easier that task will be.

But then again, the work has to be done at some point anyhow, so I'd probably merge your PRs if you do it.

> > I'm considering redoing `header-translator` the "proper" way, which is to use the existing code that Swift uses, see #345 - so possibly CoreFoundation-like structs will only happen after that.
> 
> That sounds like a really cool idea actually!

Yeah, and also quite daunting since I know so little C++ 😁

---

Relatedly, I'm feeling a tinge of burnout, so I'll probably try to do something other than `objc2` (or just not touch my computer at all) for the next few days, so if I'm a bit slow to answer that's why ;)

PR #359, opened 2023-01-18
Sendability attributes

madsmtm:
Swift recently implemented a better concept of sendability, see `NS_SWIFT_SENDABLE` and `NS_SWIFT_UI_ACTOR` in `Foundation/NSObjCRuntime.h`.

Using these attributes, we can automatically mark `NSError`, `NSDate`, `NSFileHandle` and so on `Send + Sync`. (That `NSLock` and subclasses are marked sendable seems a bit like a bug? But it shouldn't matter much, since the methods on it is `unsafe` anyhow, and deallocation on other threads is definitely safe).

More importantly, we can automatically mark subclasses of `NSResponder` (`NSView`, `NSWindow`, ...) as `MainThreadOnly`.

---

To finalize the safety of that, it would be nice to run an algorithm like this on every function/method:
```rust
if method.return_type.includes_ns_swift_ui_actor() {
    if method.receiver.is_ns_swift_ui_actor() {
        break;
    }
    for argument in method.arguments {
        if argument.is_ns_swift_ui_actor() {
            break;
        }
    }
    method.arguments.push(("mtm", Ty::MainThreadMarker));
}
```

(Note here the difference between `includes_ns_swift_ui_actor` and `is_ns_swift_ui_actor`; `includes_` returns true for `NSArray<NSView>` or `Option<NSView>`, while `is_` doesn't).

So e.g. `NSView::new() -> Id<Self>` would become `NSView::new(_mtm: MainThreadMarker)`, while `NSWindow::view(&self) -> Id<NSView>` would stay as-is.

---

I checked if we could make `atomic` attributes safe to access from any thread, and it turns out, we cannot, Swift still requires an `await` around these, so we must as well (can be tested on https://online.swiftplayground.run/):
```swift
import AppKit

@MainActor func test(window: NSWindow) {
    if #available(macOS 10.15, *) {
        Task.detached {
            // Call method; should definitely not be possible without `await`.
            await window.setFrame(NSRect.zero, display: false)

            // Access otherwise `atomic` property; turns out that is not possible without `await` either.
            print(await window.title)

            // Assign to property; likewise not possible without `await`.
            //
            // See https://stackoverflow.com/q/69259397 for why we wrap the assignment
            await MainActor.run {
                window.title = "test"
            }
        }
    }
}
```

If the property is marked `nonisolated` though, then it is possible.


madsmtm:
What about allocation? Should we make `NSWindow::alloc` require `MainThreadMarker`?

madsmtm:
A bit unsure where autogenerated `_mtm: MainThreadMarker` arguments would be most convenient to put? As the first argument (after `self`, if present), or as the last one?

I'll need to dig up some good examples to see what would be nicest?

madsmtm:
We still need to figure out how to handle `NonIsolated` methods, probably have to do something like implement them multiple times, once for `MyClass`, and once for `MainThreadBound<MyClass>`.

PR #360, opened 2023-01-19
Add webkit example

silvanshade:
I created a small example using the generated WebKit bindings. If you have any ideas on how to improve the structure, feel free to make whatever changes you'd like.

I also made some feature sets for the examples so that they can be run a little easier and I added a README.


silvanshade:
I also had some observations about the example that I thought I would mention to get your feedback about.

I wasn't quite sure how to properly define the init function so I basically followed how it was done in the `delegate.rs` example:

```rust
        #[method(initWithTextField:andWebView:)]
         #[allow(non_snake_case)]
         unsafe fn initWith(self: &mut Self, text_field: &NSTextField, web_view: &WKWebView) -> Option<&mut Self> {
             let this: Option<&mut Self> = msg_send![super(self), init];
             this.map(|this| {
                 *this.text_field = Id::retain(text_field as *const _ as *mut _).unwrap();
                 *this.web_view = Id::retain(web_view as *const _ as *mut _).unwrap();
                 this
             })
         }
```

First, I was a little surprised at the type given for this method. I would have expected it to look closer to the init methods defined in the icrate generated code, so something more like:

```rust
fn initWith(this: Option<Allocated<Self>>, text_field: &NSTextField, web_view: &WKWebView) -> Id<Self, Shared>;
```

But it seems there are a couple problems with this. One is that, if I take `this: Option<Allocated<Self>>` as an argument, I don't see any direct way to obtain an `Id<...>` from that.

If [`Allocated::option_into_ptr`](https://github.com/madsmtm/objc2/blob/8977bd1b468a652c2e3b74c472428f94afb777a5/crates/objc2/src/rc/allocated.rs#L67) were public, I could call that and then `Id::retain` to get an `Id<Self, Shared>`. But even in that case (if `Allocated::option_into_ptr` were public), I would still find it a little surprising to need to deal with raw pointers just to define `init` or `new`. It's also surprising to even need to use `msg_send` for this part since I can already call `Self::alloc` directly.

Secondly, this part feels like a hack:

```rust
this.map(|this| {
    *this.text_field = Id::retain(text_field as *const _ as *mut _).unwrap();
    *this.web_view = Id::retain(web_view as *const _ as *mut _).unwrap();
    this
})
```

I think I could have just passed the arguments `text_field` and `web_view` as `Id<...>` and cloned them at the call site, but then the signature would have again looked different to the ones in the generated code (e.g., taking `Id<...>` rather than references).

I get the impression I am either missing some part of the API that would facilitate doing these things in a cleaner way or perhaps there is some functionality missing from the public interface that isn't in place yet.

Any thoughts about that?

crates/icrate/examples/README.md
```diff
@@ -0,0 +1,8 @@
+#### Running the examples
+
+The examples can be run as follows:
+
+```sh
+cargo run --release -p icrate --example basic_usage --features=example-basic_usage
+cargo run --release -p icrate --example delegate --features=example-delegate
```
madsmtm, line 7:
```suggestion
cargo run --package=icrate --example=basic_usage --features=example-basic_usage
cargo run --package=icrate --example=delegate --features=example-delegate
...
```

crates/icrate/Cargo.toml
```diff
@@ -117,6 +111,54 @@ unstable-private = []
 # https://github.com/madsmtm/objc2/issues/new
 unstable-static-nsstring = []
 
+# Examples
```
madsmtm, line 114: I think I'd like these to have the prefix `unstable-` to clearly signify that they're not part of the stable contract. Or perhaps just have them placed in a different crate altogether?

crates/icrate/examples/browser.rs
```diff
@@ -0,0 +1,284 @@
+type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
+type BoxResult<T> = Result<T, BoxError>;
```
madsmtm, line 2: Nit: Put these after imports

crates/icrate/examples/browser.rs
```diff
@@ -0,0 +1,284 @@
...
+                Id::retain(text_field).and_then(|text_field| {
+                    Id::retain(web_view).and_then(|web_view| {
+                        Ivar::write(&mut this.text_field, text_field);
+                        Ivar::write(&mut this.web_view, web_view);
+                        Some(this)
+                    })
+                })
+            })
+        }
+
```
madsmtm, line 52: Add something similar to `unsafe impl ConformsTo<NSControlTextEditingDelegate> for AppDelegate {` here

crates/icrate/examples/browser.rs
```diff
@@ -0,0 +1,284 @@
+type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
madsmtm, line 1: Nit: Add `#![deny(unsafe_op_in_unsafe_fn)]`. I get that it's tempting to do `unsafe fn run()`, but I'd much rather actually mark the methods we're using here as safe inside `header-translator/src/data`

madsmtm (changes requested):
Yes please, examples are very much welcome!

crates/icrate/Cargo.toml
silvanshade: Sure, that's a good idea. I changed these to `unstable-` prefix for now. Personally, I think it's probably fine to keep them in the same crate but I don't really have a strong opinion about it either way so whatever you think is best.

silvanshade:
One observation I had about writing the init method is that it would be nice if we could somehow use `?` but I wasn't sure if it was possible to return a `Result` here:

```rust
unsafe fn __init_withTextField_andWebView(self: &mut Self, text_field: *mut NSTextField, web_view: *mut WKWebView) -> Option<&mut Self> {
    let this: Option<&mut Self> = msg_send![super(self), init];
    this.and_then(|this| {
        Id::retain(text_field).and_then(|text_field| {
            Id::retain(web_view).and_then(|web_view| {
                Ivar::write(&mut this.text_field, text_field);
                Ivar::write(&mut this.web_view, web_view);
                Some(this)
            })
        })
    })
}
```

The `.and_then` chaining feels a little clunky though. Ideally `Id::retain` would return a `Result` there too then, or maybe we could have some variant of it that does (since writing `.ok_or(<some long-ish reason>)` repeatedly is still verbose IMO).

If [`try_blocks`](https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html) were stable, that would also work nicely here if returning `Result` is generally problematic, since we could just write a try-block and terminate with `.ok()`.

madsmtm:
> One observation I had about writing the init method is that it would be nice if we could somehow use `?`

You can use `?` on `Option<T>`.

After https://github.com/madsmtm/objc2/pull/173, you would even be able to write:

```rust
#[method_id(initWithTextField:andWebView:)]
unsafe fn __init_withTextField_andWebView(this: Allocated<Self>, text_field: *mut NSTextField, web_view: *mut WKWebView) -> Option<Id<Self, Owned>> {
    let this = unsafe { msg_send_id![super(self), init] }?;
    Ivar::write(&mut this.text_field, unsafe { Id::retain(text_field) }?);
    Ivar::write(&mut this.web_view, unsafe { Id::retain(web_view) }?);
    Some(this)
}
```

Also, I have ideas for making `Id::retain` safe for some cases, see https://github.com/madsmtm/objc2/issues/399.

crates/icrate/examples/browser.rs
madsmtm: I actually meant for you to move the `ConformsTo` stuff _inside_ `declare_class!`, since that allows it to ensure that you've implemented any required methods, as well as ensuring that the methods are implemented correctly!

madsmtm (changes requested):
Thanks for doing the `unsafe` change; I know it looks ugly, but hopefully with https://github.com/madsmtm/objc2/pull/359 and the property stuff in https://github.com/madsmtm/objc2/issues/330, things should get a lot better!

crates/icrate/examples/browser.rs
```diff
@@ -0,0 +1,301 @@
...
+                flag,
+            )
+        }
+    };
+
+    // create the web view
+    let web_view = {
+        let this = WKWebView::alloc();
+        // let frame = window.frame();
+        let frame_rect = NSRect::ZERO;
```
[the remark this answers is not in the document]
silvanshade: I think this should be okay actually. All of the rects for the other views are also set to zero. As far as I understand it, the NSStackView handles resizing the frames appropriately when the views are added as arranged subviews.

But I'm not a macos UI expert so if there's some reason this is a problem I can change it.

(I did mean to remove the old commented out part though)

silvanshade:
> You can use `?` on `Option<T>`.

Very interesting. Did this change recently? I thought it used to not work (or it kind of did, but there was some problem involving `NoneError`)?

crates/icrate/examples/browser.rs
```diff
@@ -0,0 +1,313 @@
...
+        ) -> Id<Self, Shared>;
+    }
+);
+
+fn main() -> BoxResult<()> {
+    run()?;
+    Ok(())
+}
+
+fn run() -> BoxResult<()> {
```
madsmtm, line 136: Nit:
```suggestion
```

madsmtm (changes requested):

madsmtm:
> > You can use `?` on `Option<T>`.
> 
> Very interesting. Did this change recently? I thought it used to not work (or it kind of did, but there was some problem involving `NoneError`)?

Well, recently is subjective, but [godbolt shows it worked since Rust 1.22](https://rust.godbolt.org/z/8aGM57516) ;)

.github/workflows/ci.yml
```diff
@@ -308,6 +308,39 @@ jobs:
...
+
+    - name: Check example `delegate`
+      run: cargo check --package=icrate --example=delegate --features=unstable-example-delegate
+
+    - name: Check example `nspasteboard`
+      run: cargo check --package=icrate --example=nspasteboard --features=unstable-example-nspasteboard
+
+    - name: Check example `speech_synthesis`
+      run: cargo check --package=icrate --example=speech_synthesis --features=unstable-example-speech_synthesis
+
```
madsmtm, line 343: I think adding the example feature to `unstable-frameworks-{whereever it works}` in `icrate/Cargo.toml` should be enough, don't need to change the CI here?

madsmtm (changes requested):

.github/workflows/ci.yml
silvanshade: > I think adding the example feature to `unstable-frameworks-{whereever it works}` in `icrate/Cargo.toml` should be enough, don't need to change the CI here?

Actually, maybe I'm misunderstanding but aren't the example features essentially covered already in that sense due to the `*_all` features? So for the WebKit example, `WebKit_all` is already included in `unstable-frameworks-*`.

In that case, wouldn't adding the example features be redundant?

Don't we also want to test that the examples actually build too?

madsmtm: Well, every feature that the examples need is enabled, _except_ for the `unstable-example-...` feature. Cargo doesn't know that that feature is not actually required to compile the examples, so it doesn't try.

> Don't we also want to test that the examples actually build too?

I'm pretty sure that `cargo test` in `Test macOS 12` build all the examples as well (see e.g. [the latest build on `master`](https://github.com/madsmtm/objc2/actions/runs/4036432822/jobs/6939026601), search for `example`)

silvanshade: Ah, okay, I see what you mean now.

madsmtm (approved):
Cool, thanks!

PR #361, opened 2023-01-20
Fix errors in test-assembly

silvanshade:
It seems these should be feature gated now otherwise they show errors.


crates/test-assembly/crates/test_ns_string/lib.rs
```diff
@@ -6,14 +6,14 @@ use icrate::Foundation::NSString;
 
 // Temporary to allow testing putting string references in statics.
 // This doesn't yet compile on other platforms, but could in the future!
-#[cfg(feature = "apple")]
+#[cfg(all(feature = "apple", feature = "unstable-static-nsstring"))]
```
madsmtm, line 9: Use `"assembly-features"` instead, since `"unstable-static-nsstring"` is not defined in `test_ns_string/Cargo.toml`

madsmtm:
Thanks!

PR #362, opened 2023-01-22
Add fix for WKNavigationAction::sourceFrame

silvanshade:
Just a small fix for the WebKit bindings which I noticed while working on the browser example.

### DIFF
--- a/crates/header-translator/translation-config.toml
+++ b/crates/header-translator/translation-config.toml
@@ -1090,6 +1090,10 @@ skipped = true
 [class.MTKTextureLoader.methods.newTexturesWithContentsOfURLs_options_error]
 skipped = true
 
+[class.WKNavigationAction.methods]
+# Not marked as nullable but should be; manually defined in fixes
+sourceFrame = { skipped = true }
+
 [class.WKWebView.methods]
 # uses `SecTrustRef` from Security.framework (not yet translated)
 serverTrust = { skipped = true }
--- a/crates/icrate/src/WebKit/fixes.rs
+++ b/crates/icrate/src/WebKit/fixes.rs
@@ -1,0 +1,11 @@
+use crate::common::*;
+use crate::WebKit::*;
+
+extern_methods!(
+    #[cfg(feature = "WebKit_WKNavigationAction")]
+    unsafe impl WKNavigationAction {
+        #[cfg(feature = "WebKit_WKFrameInfo")]
+        #[method_id(@__retain_semantics Other sourceFrame)]
+        pub unsafe fn sourceFrame(&self) -> Option<Id<WKFrameInfo, Shared>>;
+    }
+);
--- a/crates/icrate/src/WebKit/mod.rs
+++ b/crates/icrate/src/WebKit/mod.rs
@@ -1,6 +1,9 @@
+mod fixes;
 #[path = "../generated/WebKit/mod.rs"]
 mod generated;
 
+#[allow(unreachable_pub)]
+pub use self::fixes::*;
 pub use self::generated::*;
 
 #[link(name = "WebKit", kind = "framework")]


madsmtm (commented):
Can you give an example of how this can fail?

Also, is this an issue in Swift as well, then?

silvanshade:
> Huh, can you give an example of how this can fail?

When I last tested it, it was actually returning null from the first call to `loadRequest` [here](https://github.com/madsmtm/objc2/blob/fd56c364194efb46fb4d932acbaa40775b126449/crates/icrate/examples/browser.rs#L294).

It might also be returning null from the call to `loadRequest` in the delegate method.

I think the time it *wasn't* null was when the navigation was triggered by a mouse click event in the webframe.

> Also, is this an issue in Swift as well, then?

I'm not sure. In the Swift docs, it does not show as being nullable though.

In any case, I submitted a bug report to Apple about it but don't know if that will lead to anything.

madsmtm:
Cool, thanks for the investigation.

Please post an update if you get any response on that report, but otherwise I'll just merge this as-is (when I get my branches together)

silvanshade:
I just tried a roughly equivalent example in Swift and it doesn't seem to return null.

I went back and tried the Rust example after adding the appropriate delegate method again and it crashes immediately.

So that's weird. I haven't tried directly with objective-C so I don't know for sure if the problem happens there.

Is it possible there is a bug in how the bindings are being generated for that case somehow?

The equivalent swift program I tried is this:

```swift
import Foundation
import WebKit

class Delegate: NSObject, NSTextFieldDelegate, WKNavigationDelegate {
    let textView: NSTextField
    let webView: WKWebView
    
    init(textView: NSTextField, webView: WKWebView) {
        self.textView = textView
        self.webView = webView
    }

    func control(_ control: NSControl, textView: NSTextView, doCommandBy commandSelector: Selector) -> Bool {
        if commandSelector == #selector(NSStandardKeyBindingResponding.insertNewline(_:)) {
            if let url = URL(string: textView.string) {
                let request = URLRequest(url: url)
                self.webView.load(request)
                return true
            }
        }
        return false
    }

    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
        print(navigationAction.sourceFrame.isMainFrame)
        decisionHandler(WKNavigationActionPolicy.allow)
    }
    
    func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
        if let url = webView.url {
            self.textView.stringValue = url.absoluteString
        }
    }
}

main()

func main() {
    let app = NSApplication.shared
    
    let window = NSWindow(contentRect: NSRect(origin: CGPoint(x: 0, y: 0), size: CGSize(width: 1024, height: 768)), styleMask: [.closable, .resizable, .titled], backing: NSWindow.BackingStoreType.buffered, defer: false)
    let webView = WKWebView(frame: NSRect.zero)
    
    let navBar = NSStackView(frame: NSRect.zero)
    navBar.orientation = NSUserInterfaceLayoutOrientation.horizontal
    navBar.alignment = NSLayoutConstraint.Attribute.height
    navBar.distribution = NSStackView.Distribution.fill
    navBar.spacing = 0
    
    let navButtons = NSStackView(frame: NSRect.zero)
    navButtons.orientation = NSUserInterfaceLayoutOrientation.horizontal
    navButtons.alignment = NSLayoutConstraint.Attribute.height
    navButtons.distribution = NSStackView.Distribution.fillEqually
    navButtons.spacing = 0
    
    let backButton = NSButton(title: "back", target: webView, action: #selector(WKWebView.goBack(_:)))
    backButton.bezelStyle = NSButton.BezelStyle.shadowlessSquare
    
    let forwardButton = NSButton(title: "forward", target: webView, action: #selector(WKWebView.goForward(_:)))
    forwardButton.bezelStyle = NSButton.BezelStyle.shadowlessSquare
    
    navButtons.addArrangedSubview(backButton)
    navButtons.addArrangedSubview(forwardButton)

    let navURL = NSTextField(frame: NSRect.zero)
    navURL.drawsBackground = true
    navURL.backgroundColor = NSColor.lightGray
    navURL.textColor = NSColor.black
    
    navBar.addArrangedSubview(navButtons)
    navBar.addArrangedSubview(navURL)
    
    let contentView = NSStackView(frame: window.frame)
    contentView.orientation = NSUserInterfaceLayoutOrientation.vertical
    contentView.alignment = NSLayoutConstraint.Attribute.width
    contentView.distribution = NSStackView.Distribution.fill
    contentView.spacing = 0
    
    contentView.addArrangedSubview(navBar)
    contentView.addArrangedSubview(webView)
    
    let delegate = Delegate(textView: navURL, webView: webView)

    navURL.delegate = delegate
    webView.navigationDelegate = delegate

    window.contentView = contentView
    window.center()
    window.title = "browser example"
    window.makeKeyAndOrderFront(nil)
        
    let request = URLRequest(url: URL(string: "https://google.com")!)
    webView.load(request)
    
    app.run()
}
```

madsmtm:
I'm pretty sure the bindings are correct, and the headers wrong.

I looked at the Swift assembly, it does roughly the same thing, and if you try:
```rust
// Print
print(Unmanaged.passUnretained(navigationAction.sourceFrame).toOpaque())
print(navigationAction.sourceFrame)
```

You will see that `sourceFrame` is indeed a NULL pointer.

The reason that `navigationAction.sourceFrame.isMainFrame` works is that Objectice-C has this (anti-)feature where message sends to NULL effectively end up returning `nil`. How does that work for other types you ask? Well, depending on the ABI, that means that the `BOOL` `isMainFrame` is interpreted as `false`.

madsmtm:
Thanks for the rebase!

PR #370, opened 2023-01-22
Add GameKit framework

silvanshade:


crates/header-translator/translation-config.toml
```diff
@@ -851,6 +879,11 @@ skipped = true
 [fn.MTLPackedFloat3Make]
 skipped = true
 
+# Needs `JSContext` from JavaScriptCore
+[class.WebFrame.methods.javaScriptContext]
+skipped = true
+[protocol.WebFrameLoadDelegate.methods.webView_didCreateJavaScriptContext_forFrame]
+skipped = true
```
silvanshade, line 886: These should have been skipped in the earlier PR adding the WebKit framework but I didn't notice them until now. For some reason, translating `GameKit` tries to change the `WebKit_JSContext` feature to `SceneKit_JSContext` (which I saw in the diff) so I just added the skips in this PR.

PR #386, opened 2023-01-23
Add GameController framework

silvanshade:


crates/icrate/src/GameController/fixes/mod.rs
```diff
@@ -0,0 +1,86 @@
+use crate::common::c_long;
+use crate::Foundation::NSString;
+
+pub type GCKeyCode = c_long;
```
madsmtm, line 4: Add note about `CFIndex` being `c_long_long` on `__LLP64__` / Windows 64-bit (and how this doesn't matter for us)

madsmtm (approved):
Thanks a lot, great work!

If you fix the CI on this PR and rename it, I'll just merge it in one go (sorry for making you go through the trouble of creating separate PRs for each framework, I thought they would be much more troublesome to review than they have been)

silvanshade:
> If you fix the CI on this PR and rename it, I'll just merge it in one go

So I'm not quite sure how to fix this one.

The problem is that the macro expands to this:

```rust
#[repr(C)]
#[derive(Clone, Copy, Debug, PartialEq)]
#[deprecated = "GCExtendedGamepadSnapshot has been deprecated, use [GCController controllerWithExtendedGamepad] instead"]
pub struct GCExtendedGamepadSnapshotData {
    pub version: u16,
    pub size: u16,
    pub dpadX: c_float,
    pub dpadY: c_float,
    pub buttonA: c_float,
    pub buttonB: c_float,
    pub buttonX: c_float,
    pub buttonY: c_float,
    pub leftShoulder: c_float,
    pub rightShoulder: c_float,
    pub leftThumbstickX: c_float,
    pub leftThumbstickY: c_float,
    pub rightThumbstickX: c_float,
    pub rightThumbstickY: c_float,
    pub leftTrigger: c_float,
    pub rightTrigger: c_float,
    // problems below
    pub supportsClickableThumbsticks: Bool,
    pub leftThumbstickButton: Bool,
    pub rightThumbstickButton: Bool,
}
```

But the last three fields have type `Bool` which doesn't implement `PartialEq`, hence the errors about `==` and `!=`.

I was going to look at adding a `PartialEq` impl for `Bool`, but then saw this comment:

```rust
[repr(transparent)]
// We don't implement comparison traits because they could be implemented with
// two slightly different semantics:
// - `self.as_bool().cmp(other.as_bool())`
// - `self.value.cmp(other.value)`
// And it is not immediately clear for users which one was chosen.
#[derive(Copy, Clone, Default)]
pub struct Bool {
    value: objc_sys::BOOL,
}
```

So I'm not sure what the best approach here would be. I could simply manually define the struct in the `fixes` mod without the `PartialEq` derive. Alternatively, maybe it would be a good idea to modify the macro to allow that derive to be optional or something.

What do you think?

> (sorry for making you go through the trouble of creating separate PRs for each framework, I thought they would be much more troublesome to review than they have been)

No worries. It was actually probably good to do it this way to start with since I wasn't really sure how it was going to work out trying to translate a lot of these frameworks. But maybe I will try to do more in chunks for future PRS or something so it's less noisy in the repo.

silvanshade:
I just opted to manually define the struct for now since that seems like the easiest solution.

madsmtm:
> good idea to modify the macro to allow that derive to be optional or something.

Yeah, we could do that at some point - but the manual fix is fine for now!

PR #396, opened 2023-01-24
Add LocalAuthentication and LocalAuthenticationEmbeddedUI frameworks

silvanshade:


crates/icrate/src/LocalAuthentication/fixes/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+        LAErrorBiometryNotAvailable = -6,
+        LAErrorBiometryNotEnrolled = -7,
+        LAErrorBiometryLockout = -8,
+        LAErrorNotInteractive = -1004,
+        LAErrorWatchNotAvailable = -11,
+        LAErrorBiometryNotPaired = -12,
+        LAErrorBiometryDisconnected = -13,
+        LAErrorInvalidDimensions = -14,
+    }
+);
```
silvanshade, line 36: The issue here seems to be that it is parsing the negative values incorrectly.

madsmtm: Yeah, things referencing defined constants like `kLAErrorTouchIDNotAvailable` is indeed a bit harder to parse correctly :/

PR #397, opened 2023-01-24
Add MailKit framework

silvanshade:


crates/header-translator/src/rust_type.rs
```diff
@@ -1672,22 +1672,22 @@ impl fmt::Display for Ty {
...
-                            id,
                             params: TypeParams::Empty,
                             ownership: None,
+                            ..
                         },
                     is_const: _,
                     lifetime: _,
                     nullability: Nullability::Nullable | Nullability::Unspecified,
-                } if id.is_nsstring() || id.is_nsdata() => {
+                } => {
```
madsmtm, line 1682: I suspect I'm going to go back on this a bit at some point (or, at least special-case `NSString`, since it would be nice to make safe), but for now it's fine

madsmtm (approved):
Looks good, thanks!

Issue #387, opened 2023-01-23
[panic] `header-translator` generates named structs for anonymous structs


tjamaan:
# Problem
Currently `header-translator` generates named structs for anonymous structs. This causes `panic!` when verifying messages:
`thread '<unnamed>' panicked at 'invalid message send to -[NSProcessInfo operatingSystemVersion]: expected return to have type code '{?=qqq}', but found '{NSOperatingSystemVersion=qqq}'',
`

# Cause
The objc type definition is as below. Notice that it's just a typedef and the struct itself doesn't have a name.
```objc
/* NSProcessInfo.h */
typedef struct /* no name */ {
    NSInteger majorVersion;
    NSInteger minorVersion;
    NSInteger patchVersion;
} NSOperatingSystemVersion;

// @encode(NSOperatingSystemVersion) = {?=qqq}
```

Here is the generated Rust code:
```rs
extern_struct!(
    pub struct NSOperatingSystemVersion {
        pub majorVersion: NSInteger,
        pub minorVersion: NSInteger,
        pub patchVersion: NSInteger,
    }
);
```

Which expands to:
```rs
#[repr(C)]
#[derive(Clone, Copy, Debug, PartialEq)]
pub struct NSOperatingSystemVersion {
    pub majorVersion: NSInteger,
    pub minorVersion: NSInteger,
    pub patchVersion: NSInteger,
}
#[cfg(feature = "objective-c")]
unsafe impl objc2::Encode for NSOperatingSystemVersion {
    const ENCODING: objc2::Encoding = (objc2::Encoding::Struct(
        "NSOperatingSystemVersion", // <--- FIXME: the struct should be anonymous
        &[
            <NSInteger as objc2::Encode>::ENCODING,
            <NSInteger as objc2::Encode>::ENCODING,
            <NSInteger as objc2::Encode>::ENCODING,
        ],
    ));
}
#[cfg(feature = "objective-c")]
unsafe impl objc2::RefEncode for NSOperatingSystemVersion {
    const ENCODING_REF: objc2::Encoding =
        objc2::Encoding::Pointer(&<Self as objc2::Encode>::ENCODING);
}
```

# Next steps
Mark anonymous structs and have a variant in `extern_struct!` that uses `"?"` as the name for the `obj2::Encoding::Struct` encoding.

tjamaan:
FYI I'm working on a patch for this.

tjamaan:
[...]

PR #398, opened 2023-01-25
Separate Rust struct type names and Objc internal struct names during header translations

tjamaan:
Fixes: #387 

Currently `header-translator` uses the struct typedef name as the struct name which is incorrect.
When Objc encodes structs it uses the struct name without any typedefs. E.g. the following struct has the internal struct name of `foo` and `@encode(bar) = {foo=q}`:
```
typedef struct foo {
    NSInteger x;
} bar;
```

This pull request fixes the header translator to store the internal struct name separately from the potentially typedef'd identifier. Additionally the `extern_struct!` macro is modified to require a `#[struct_name("name")]` attribute to generate the proper `Encoding`.

# Next step
Rerun `header-translator` and update [`icrate-generated`](https://github.com/madsmtm/icrate-generated/).

### DIFF
--- a/crates/header-translator/src/data/Foundation.rs
+++ b/crates/header-translator/src/data/Foundation.rs
@@ -127,6 +127,7 @@ data! {
     class NSProcessInfo {
         unsafe +processInfo;
         unsafe -processName;
+        unsafe -operatingSystemVersion;
     }
 
     class NSSet {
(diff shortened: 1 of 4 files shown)


tjamaan:
The failing tests will be fixed once `header-translator` is rerun and `icrate-generated` is updated.

madsmtm (approved):
Thanks! I'll merge this at some point when I get my other branches in order

PR #400, opened 2023-01-27
Add CoreFoundation framework

silvanshade:
I thought I would give `CoreFramework` a try finally since so many of the remaining frameworks rely on it one way or another.

I opted to add the `libc` crate as a dependency to define some of the things needed. This may be relevant for #305.


samuelsleight:
What's the current status of this branch/pull request? I'm looking into adding support for the `AVFoundation` framework, which I think makes sense to base off this work to add `CoreFoundation` given it looks like it pulls in a lot of other frameworks that all also depend on this.

I _do_ have a (currently unpushed) branch with the conflicts resolved on this that I'd made _some_ progress on, and I'm happy to look into tidying that up at some point, if this current PR is essentially dead?

samuelsleight:
Fair enough, I'll see where I get to with that then!

I'll likely also submit a PR for at least one other framework beforehand, as there's at least one that doesn't appear to depend on this or indeed anything else

PR #407, opened 2023-01-29
Add MetricKit framework

silvanshade:


crates/icrate/src/MetricKit/fixes/mod.rs
```diff
@@ -0,0 +1,33 @@
+use crate::common::*;
+use crate::Foundation::*;
+use crate::MetricKit::*;
+
+pub type MXLaunchTaskID = NSString;
+
+// pub type os_log_t = ProtocolObject<dyn OS_os_log>;
+
+// extern_protocol!(
+//     pub unsafe trait OS_os_log: NSObjectProtocol {}
```
madsmtm, line 10: Yeah, this is an undocumented protocol, so not sure what we may rely on there.

Issue #393, opened 2023-01-24
List of frameworks to translate


silvanshade:
This is a list of frameworks I've been working through in the translation PRs.

This is basically just what shows up on the Apple website under system frameworks, technologies page, the SDK directories, etc.

Some of these are obsolete and probably not worth bothering with. Several have issues that will prevent translation until more work is done on the translator (e.g., the core foundation stuff). Some give errors related to seemingly not finding the header files (not sure why, but I've added notes about those below).

Some are only available in Swift and I guess this will only increase in the future. It would be nice if we could figure out a way to still include bindings for those but I suppose it might require building out some new tooling?

- [ ] `Accelerate`
- [x] `Accessibility`
- [ ] `Accounts` (deprecated)
- [ ] `AddressBook` + `AddressBookUI.h`
- [ ] `ActivityKit` (swift)
- [x] `AdServices`
- [x] `AdSupport`
- [ ] `AGL` (deprecated)
- [ ] `AppClip`
- [ ] `AppIntents` (swift)
- [x] `AppKit`
- [ ] `AppleArchive`
- [ ] `AppleCryptoKit` (swift)
- [ ] `AppleScriptKit`
- [ ] `AppleScriptObjC`
- [ ] `ApplicationServices` (needs Carbon stuff)
- [x] `AppTrackingTransparency` (doesn't find header?)
- [ ] `ARKit` (needs iOS?)
- [ ] `AssetsLibrary` (deprecated, needs iOS)
- [ ] `AudioDriverKit`
- [ ] `AudioToolbox`
- [ ] `AudioUnit`
- [ ] `AudioVideoBridging`
- [x] `AuthenticationServices`
- [x] `AutomaticAssessmentConfiguration`
- [x] `Automator`
- [ ] `AVFAudio`
- [ ] `AVFKit`
- [ ] `AVFoundation`
- [ ] `AVRouting`
- [x] `BackgroundAssets`
- [x] `BackgroundTasks`
- [ ] `BlockStorageDeviceDriverKit`
- [x] `BusinessChat`
- [ ] `CalendarStore` (deprecated?)
- [x] `CallKit`
- [ ] `Carbon`
- [ ] `CareKit`
- [ ] `CarKey` (requires special entitlement)
- [ ] `CarPlay` (catalyst / iOS)
- [ ] `CFNetwork`
- [ ] `CHIP`
- [x] `ClassKit`
- [ ] `ClockKit` (catalyst)
- [x] `CloudKit`
[...]

Issue #310, opened 2022-12-23
Fix enums and typedefs in `icrate`


madsmtm:
Right now we're just using type aliases - instead, we should generate newtypes with the appropriate traits and constants.

silvanshade:
I believe the implementation I've been working on in #425 does a reasonably good job now of handling enums at least.

I generally tried to follow a similar approach for mapping things over to how the enums are translated for Swift. There are some caveats, for example due to the fact that we can't define an `enum` within a `struct` in Rust like you can in Swift, or not having `impl` for `enum`, or not (yet) having associated types for inherent impls.

The most sophisticated example is `icrate/examples/ns_error_enum.rs` where I use the macro to create an error enum translation for `NSURLError`. (See the code snippet and macro expansion at the end for reference).

A couple things to note:

You can (optionally) specify the name of the helper module with `mod`. I think it might be reasonable to update the macro with a list of Apple system framework prefixes and special cases so this isn't needed (the only reason it's used here is for correct tokenization of `NSURL`). 

You can also (optionally) specify refinements for the `userInfo` dictionary for the error. The syntax is `{ key = <ident>, type = <type>, from? = <expr>, into? = <expr> }`. The `key` is expected to be an identifier in scope that resolves to a dictionary key. The `type` is the type of the associated field in the generated `UserInfo` struct (which is then wrapped in `Option<...>`). The optional `from` and `into` keys are used to convert the entry _from_ the `NSDictionary` to the `UserInfo` field, and then back from the field _into_ the entry.

Similar to Swift, getters are generated for the specified `UserInfo` fields, but you can also pattern match on the entire `UserInfo` struct (which uses the getters to populate) like so:

```rust
// `UserInfo<T>` is an alias for `<T as objc2::ErrorEnum>::UserInfo` (to avoid mentioning the helper module)
let UserInfo::<NSURLError> {
    failing_url,
    failing_url_string,
    background_task_cancelled_reason,
    network_unavailable_reason,
} = error.user_info();
let failing_url_string = error.failing_url_string();
```

The error declaration looks like this:

```rust
#[objc(error_enum,
    mod = ns_url_error,
    domain = unsafe { NSURLErrorDomain },
    user_info = [
        // NOTE: Here we specify (typed) getters for the associated `userInfo` dict.
        { key = NSURLErrorFailingURLErrorKey, type = Id<NSURL> },
        { key = NSURLErrorFailingURLStringErrorKey, type = Id<NSString> },
[...]

Issue #55, opened 2021-10-30
My take on `derive` macros


madsmtm:
We have a few traits that users has to implement manually, namely `Encode`, `RefEncode`, `Message` and the `INS...` traits. It would be nice if these could be automatically `derive`d, but in essence we can't allow that because they are `unsafe`, and a `derive` macro wouldn't be able to uphold the required safety invariants.

`#[derive(Encode)]` (requested upstream in https://github.com/SSheldon/rust-objc/issues/52) specifically doesn't work because the user could create a `#[repr(C)]` struct which used e.g. `[i8; 4]` while the Objective-C side used `int`. Or `*const u8` vs. `c_void*`; this is valid from a normal FFI-perspective, but `Encode` requires that the encodings match perfectly.

So I'm actually probably going to remove the existing `objc2_foundation_derive` helpers in this repo.

madsmtm:
> a `derive` macro wouldn't be able to uphold the required safety invariants.

> Or `*const u8` vs. `c_void*`; this is valid from a normal FFI-perspective, but `Encode` requires that the encodings match perfectly.

I'm finding, when converting huge amounts of code, that derived `Encode` and `RefEncode` might actually be useful... Maybe we can relax the safety requirements in them so that they don't have to be implemented correctly, only that they have to be correct FFI-wise? Or something...
Will have to match the `objc_msgSend` functions in `message/apple/...`!

madsmtm:
We could maybe also add a `"old-compat"` feature flag, which also implements `Encode` from the `objc v0.2.7` crate; then crates from e.g. `core-foundation-rs` wouldn't need to choose between `objc` and `objc2` support:
```rust
unsafe impl objc::Encode for T {
    #[inline]
    fn encode() -> objc::Encoding {
        objc::Encoding::from_str(T::ENCODING)
    }
}
```

PR #411, opened 2023-01-29
Allow ivar names to be elided

silvanshade:
This PR makes the ivar names optional (re: my earlier comment).

I attempted to do the same for the ivar helper module but it seemed problematic without more serious refactoring of the macros. The reason being that you wouldn't want to create the helper module if there were no ivars specified, but this is hard to detect without more intermediate macros.

One thought I had though is that it feels like many of the macros are rather complicated. I wonder if it might not be worth considering rewriting many of them as proc-macros.

There are the usual downsides to that approach, but on the other hand, it would make some things much easier to implement (like allowing the ivar helper module to be optional) and it would allow us much more control over what kind of error messages we could produce.

I also wonder, especially given the size of `icrate`, whether it might make an impact in terms of performance. As I note in #393, there's at least one framework (`Matter`) that I couldn't get to finish compiling even after raising the recursion limit significantly (up to 2048). In that case I think the issue was partly due to a giant enum, which with a proc macro, we could handle with iteration rather than recursion.

I realize it would be quite a lot of work to rewrite those but that's something I would be willing to help with if you think it's a good idea.

### DIFF
--- a/crates/icrate/examples/delegate.rs
+++ b/crates/icrate/examples/delegate.rs
@@ -14,12 +14,12 @@ extern "C" {}
 declare_class!(
     #[derive(Debug)]
     struct CustomAppDelegate {
-        pub ivar: IvarEncode<u8, "_ivar">,
-        another_ivar: IvarBool<"_another_ivar">,
-        box_ivar: IvarDrop<Box<i32>, "_box_ivar">,
-        maybe_box_ivar: IvarDrop<Option<Box<i32>>, "_maybe_box_ivar">,
-        id_ivar: IvarDrop<Id<NSString, Shared>, "_id_ivar">,
-        maybe_id_ivar: IvarDrop<Option<Id<NSString, Shared>>, "_maybe_id_ivar">,
+        pub ivar: IvarEncode<u8>,
+        another_ivar: IvarBool,
+        box_ivar: IvarDrop<Box<i32>>,
+        maybe_box_ivar: IvarDrop<Option<Box<i32>>>,
+        id_ivar: IvarDrop<Id<NSString, Shared>>,
+        maybe_id_ivar: IvarDrop<Option<Id<NSString, Shared>>>,
     }
 
     mod ivars;
--- a/crates/objc2/src/macros/__field_helpers.rs
+++ b/crates/objc2/src/macros/__field_helpers.rs
@@ -180,7 +180,7 @@ macro_rules! __parse_fields {
     (
         (
             $(#[$m:meta])*
-            $vis:vis $field_name:ident: IvarDrop<$ty:ty, $ivar_name:literal>
+            $vis:vis $field_name:ident: IvarDrop<$ty:ty $(, $ivar_name:literal)?>
             $(, $($rest_fields:tt)*)?
         )
         ($($ivar_helper_module_v:vis mod $ivar_helper_module:ident)?)
@@ -208,7 +208,7 @@ macro_rules! __parse_fields {
                 // - Caller upholds that the ivars are properly initialized.
                 unsafe impl $crate::declare::IvarType for $field_name {
                     type Type = IvarDrop<$ty>;
-                    const NAME: &'static $crate::__macro_helpers::str = $ivar_name;
+                    const NAME: &'static $crate::__macro_helpers::str = $crate::__select_ivar_name!($field_name; $($ivar_name)?);
                 }
             ) ($($ivar_type_name)* $field_name)
             (
@@ -227,7 +227,7 @@ macro_rules! __parse_fields {
     (
         (
             $(#[$m:meta])*
-            $vis:vis $field_name:ident: IvarEncode<$ty:ty, $ivar_name:literal>
+            $vis:vis $field_name:ident: IvarEncode<$ty:ty $(, $ivar_name:literal)?>
             $(, $($rest_fields:tt)*)?
         )
         ($($ivar_helper_module_v:vis mod $ivar_helper_module:ident)?)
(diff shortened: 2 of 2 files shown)


madsmtm:
Linking your comment here: https://github.com/madsmtm/objc2/commit/42b953804c0b8f5bed98cd3e9be29c7d764626e4#r98502810.

> I was just wondering, but is there a rationale for needing to be able to specify these manually? I'm referring to the ivar name and the module name. Maybe they could just be created automatically in the macro invocation? Or perhaps at least allow a macro rule case where they can be elided?

For the module name, yes, they could probably be generated automatically without too much trouble by using the [`paste`](https://crates.io/crates/paste) macro or similar - but that requires proc-macros, see below.

For the instance variable name, I provided a short rationale in https://github.com/madsmtm/objc2/pull/409, but I'll try to expand on it:
1. The name must be unique across all superclasses (this is something I only just now figured out). There is no way to statically know which names are taken, and that may even change in new OS versions, so e.g. in `winit` we'll probably have to do `"winit_ivar"`.
2. The name will be present in the final binary, which is something that the user should be aware about
   - Users that want to "protect" their code / keep the binary size small are used to being able to strip debug info, and then be done with it. They will not be able to with this.
  - Users that want to give a better debugging experience would be unsatisfied with a long, auto-generated name.

But perhaps I'm being pedantic, and nobody actually cares what the name is? What do you think?

To avoid conflicts though, I'd probably prefer the name being something like `concat!("_", class_name, "_", field_name)` instead.

> One thought I had though is that it feels like many of the macros are rather complicated. I wonder if it might not be worth considering rewriting many of them as proc-macros.

When I started this project, it was a direct competitor to `objc`, which meant that the extra compile-time for a procedural macro + `syn` + `quote` was out of the question (and I dismissed e.g. `#[derive(Encode, RefEncode)]` on those grounds as well: https://github.com/madsmtm/objc2/issues/55). Though I do admitm, now that we have `icrate`, things might have changed enough to reconsider.

Another issue with procedural macros is that they're less visible to IDEs, though I'm not sure to what extent this is, and what the recommendations are for overcoming this? Will probably at least require some work to allow the macro to always emit `compile_error!(...)` instead of panicking, such that type-information is available to IDEs in most cases.

> would allow us much more control over what kind of error messages we could produce.

This is the biggest point for me, complex declarative macros have quite terrible UI.

~And e.g. `extern_methods!` could be gone altogether if we had attribute macros `method` and `method_id`~. Well, perhaps not, we'd still need a way to mark it as `unsafe`. But it could be `#[extern_methods] unsafe impl ...` instead. Would even allow putting `#[extern_methods]` inside `declare_class!`, which would save some repetition.

> I also wonder, especially given the size of `icrate`, whether it might make an impact in terms of performance. As I note in #393, there's at least one framework (`Matter`) that I couldn't get to finish compiling even after raising the recursion limit significantly (up to 2048). In that case I think the issue was partly due to a giant enum, which with a proc macro, we could handle with iteration rather than recursion.

There _are_ ways to deal with that, see [TT munchers performance](https://veykril.github.io/tlborm/decl-macros/patterns/tt-muncher.html#performance). For `extern_enum!` specifically, the current implementation was mostly just a hack, it will likely go away after https://github.com/madsmtm/objc2/issues/310.

> I realize it would be quite a lot of work to rewrite those but that's something I would be willing to help with if you think it's a good idea.

I would be very interested in a proof-of-concept of replacing e.g. `extern_methods!` with a procedural macro (as a test it doesn't have to handle `method` vs. `method_id` and such, but one that does roughly the same work (beware of emitting `todo!()` or likewise, since that will allow the compiler to do less work)), and seeing what the effect on performance would be for a `cargo build --package=icrate --features=Foundation_all` - since that's the only real holdout for me.

silvanshade:
Thanks for the details response, that clarifies a lot of things.

> But perhaps I'm being pedantic, and nobody actually cares what the name is? What do you think?

Good question.

I think if it comes down to the choice between better ergonomics for typical, non-debugging use cases (i.e., you just want to declare a class and assume everything is going to work), versus being a little easier to read in the case where things go wrong and you do need to do the debugging, I would probably favor better ergonomics in the typical case.

So in that sense, if uniqueness is the issue, I would opt for some autogenerated naming scheme even if the results are a bit unwieldy. I suppose we could still allow the names to be optional in the case that the user really cares, right?

> I would be very interested in a proof-of-concept of replacing e.g. extern_methods! with a procedural macro (as a test it doesn't have to handle method vs. method_id and such, but one that does roughly the same work (beware of emitting todo!() or likewise, since that will allow the compiler to do less work)), and seeing what the effect on performance would be for a cargo build --package=icrate --features=Foundation_all - since that's the only real holdout for me.

Okay, sounds goods. I will give that a try and we can see what the results look.

madsmtm:
> So in that sense, if uniqueness is the issue, I would opt for some autogenerated naming scheme even if the results are a bit unwieldy. I suppose we could still allow the names to be optional in the case that the user really cares, right?

Yeah, I would be fine with that (though being able to specify it manually is very useful for the case where you want to use `objc2` to interface with Rust from Objective-C, so allowing manual overriding would still be nice).

I did _intentionally_ remove the ability to specify the class name automatically, since I suspect it's mostly just causes code-smells (e.g. in `winit` I am doing `struct WinitView { ... }` and `struct WinitApplicationDelegate { ... }`, which could've been shortened a lot) (and I don't _think_ there's a way to query the current crate, such that we could autogenerate names there as well?).

madsmtm:
Looking at it again actually, I think I've been mistaken? If you use `class_getInstanceVariable` with the correct class, then you can access instance variables correctly no matter if two classes may have the same instance variable name.

In effect, I think it is just be _our_ usage of `class_getInstanceVariable` that's incorrect, we use the dynamic class from `obj.class()`, instead of the static class from `ClassType::class`, which is what causes the undesired behaviour!

(This also means that usage of `object_getInstanceVariable`/`object_setInstanceVariable`/`AnyObject::ivar`/`AnyObject::set_ivar` should be heavily discouraged, since they all inhibit this problem!)

Was going to fix this as part of https://github.com/madsmtm/objc2/issues/414 anyhow, just writing down my thoughts for now though.

PR #413, opened 2023-01-30
Proof of concept for proc macros

silvanshade:
Following the discussion from #411


crates/objc2-proc-macros/src/extern_methods.rs
```diff
@@ -2,6 +2,50 @@ use proc_macro2::TokenStream;
...
+
+// Examples:
+//
+//   #[method(sel = initTextCell:, retain = init)]
+//   pub unsafe fn initTextCell(
+//       this: Option<Allocated<Self>>,
+//       string: &NSString,
+//   ) -> Id<Self, Shared>;
+//
+//   #[method(sel = loadHTMLString:baseURL:, retain)] // retain without = defaults to `other`
```
silvanshade, line 19: We could probably make specifying even `retain` optional in the typical case since we could just determine this by looking at the return type of the signature.

crates/objc2-proc-macros/src/extern_methods.rs
```diff
@@ -2,6 +2,50 @@ use proc_macro2::TokenStream;
...
 
+enum RetainSemantics {
+    Init,
+    New,
+    Other,
+}
+
+// Examples:
+//
+//   #[method(sel = initTextCell:, retain = init)]
```
silvanshade, line 13: Similar to my other comment, we could allow eliding the retain info here as well (by examining the type of `this`).

silvanshade:
Thinking about this some more, it seems like we might be able to simplify things more than I anticipated.

What if we had something like this:

```rust
#[class(extern, super = NSActionCell, inherits = NSCell, NSObject)]
struct NSPathCell;

// `unsafe impl ClassType for NSPathCell` added automatically

#[class]
unsafe impl NSPathCell {
    // #[method] can often be skipped entirely
    pub unsafe fn pathStyle(&self) -> NSPathStyle;
    pub unsafe fn setPathStyle(&self, path_style: NSPathStyle);
    #[cfg(all(feature = "Foundation_NSArray", feature = "Foundation_NSString"))]
    pub unsafe fn allowedTypes(&self) -> Option<Id<NSArray<NSString>, Shared>>;
}

#[protocol] // `: NSObjectProtocol` can be added automatically
pub unsafe trait NSPathCellDelegate {
    #[cfg(all(feature = "AppKit_NSOpenPanel", feature = "AppKit_NSPathCell"))]
    #[method(optional)] // selector computed from name
    unsafe fn pathCell_willDisplayOpenPanel(
        &self,
        path_cell: &NSPathCell,
        open_panel: &NSOpenPanel,
    );
}

// `unsafe impl ProtocolType for dyn NSPathCellDelegate {}` added automatically
```

Anyway, I might try experimenting with a few different approaches to see what seems feasible.

madsmtm:
> // #[method] can often be skipped entirely

Don't spend too much time on this; we'll want different names for methods anyhow: https://github.com/madsmtm/objc2/issues/284

> ```
> #[class(extern, super = NSActionCell, inherits = NSCell, NSObject)]
> struct NSPathCell;
> ```

I'm not sure about this one, it would allow writing something like `#[class(super = NSString)] struct NSCell;` without `unsafe` anywhere. We'd have to make sure that case was sound (if not still nonsensical).

madsmtm:
Also, I suspect I'd like to keep `msg_send!` and `msg_send_id!` as non proc-macros; that way, if users just want a better `objc`, they can opt out of the more advanced functionality, and just get the improvements that `msg_send_id!` provides.

crates/objc2-proc-macros/src/extern_methods.rs
madsmtm: The reason that I did it the `@__retain_semantics` `extern_methods!` originally is because of performance (otherwise it uses `const` stuff to figure the selector out), but doing such work is possibly performant enough when running as a proc macro, so we might not need it at all.

madsmtm:
Superseded by https://github.com/madsmtm/objc2/pull/425

PR #416, opened 2023-02-01
Allow using `MainThreadMarker` in `extern_methods!`

madsmtm:
Required for https://github.com/madsmtm/objc2/pull/359.

Unsure if this can be supported in `declare_class!`.


madsmtm:
Alternative: Move `MainThreadMarker` into `objc2` + make some trait that allows it to be skipped as an argument in `msg_send!`

madsmtm:
Yet another alternative: Handle this outside of `extern_methods!`, and instead push the user to define their own wrapper methods.

madsmtm:
Although I don't really like it, I went with it for now, then if we get something like contexts and capabilities then we can rethink this.

Also, it may be valuable to do this for `AutoreleasePool` as well?

Issue #401, opened 2023-01-27
Fix `NSCopying` and `NSMutableCopying`


madsmtm:
In https://github.com/madsmtm/objc2/pull/293 / https://github.com/madsmtm/objc2/issues/291, I changed protocols to work like traits.

However, `NSCopying` and `NSMutableCopying` are still special-cased since they work a bit differently, they require the user to specify the output type and ownership. How can we make `icrate` do that for us? And what about protocols that "inherit" from these two (like `DOMEventTarget`)?

Perhaps this ties in with https://github.com/madsmtm/objc2/issues/399 somewhat?

In any case, I need to get a better understanding and research which types actually implement these two protocols, and just as importantly, which ones don't!

madsmtm:
Looking at the conforming types of [`NSMutableCopying`](https://developer.apple.com/documentation/foundation/nsmutablecopying?language=objc) to start with, it seems that the only classes that implement this while _not_ having a `Mutable` class counterpart is:
- [`SWCollaborationMetadata`](https://developer.apple.com/documentation/sharedwithyou/swcollaborationmetadata?language=objc)
- [`AVAudioBuffer`](https://developer.apple.com/documentation/avfaudio/avaudiobuffer?language=objc)

Everything else works just like `NSString` vs. `NSMutableString`.

madsmtm:
A bit more complex example:

| Class | `copy` | `mutableCopy` |
| --- | --- | --- |
`NSTextStorage` | `NSAttributedString` | `NSMutableAttributedString` |
`NSMutableAttributedString` | `NSAttributedString` | `NSMutableAttributedString` |
`NSAttributedString` | `NSAttributedString` | `NSMutableAttributedString` |
`NSObject` | crash | crash |

madsmtm:
So in https://github.com/madsmtm/objc2/pull/419 I was considering getting rid of the `Output` parameter, but I think that might not be desirable after all? Unsure, let's investigate:

The places where it is troublesome is in usage of `ProtocolObject<dyn T>` where `T` is a subtrait of `NSCopying`:
- `MTLFunctionStitchingFunctionNode::arguments`
- `DOMEvent::target/currentTarget/srcElement` + `DOMMouseEvent::relatedTarget`
- `ASAuthorizationPublicKeyCredentialAssertionRequest::allowedCredentials`
- `ASAuthorization::credential`
- `AXChartDescriptor::xAxis/additionalAxes`

[...]

Issue #399, opened 2023-01-26
Making `Id::retain` safer


madsmtm:
For some things like `NSMutableString` (and by extension, `NSString`), calling `Id::retain` on `&T` cannot be safe, since there could be a mutable reference to it somewhere else in the program (and the lifetime is erased by calling `Id::retain`, that's kinda the whole point).

The solution there is to use `NSCopying::copy` to instead store a copy of your string (which is cheap when given `&NSString`, and correct when given `&NSMutableString`).

But for types like `NSView` and `NSWindow`, this is overly restrictive, since we know there can never be a mutable reference to these (they always use interior mutability); so perhaps we can make a trait for expressing this? (This comes up especially often in user-defined `init` methods).

madsmtm:
Making `retain` is safe for a specific class does require that _no subclasses ever use `Id<_, Owned>`_. Which is, well, hard!

madsmtm:
This also relates to https://github.com/madsmtm/objc2/issues/401 and https://github.com/madsmtm/objc2/issues/265.

I'm starting to cook up a scheme like this:
```rust
trait ClassKind: Sealed {}

// Fallback for everything that hasn't been specified
struct Unknown;
impl ClassKind for Unknown {}

/// `Id::retain` is safe
struct Immutable;
impl ClassKind for Immutable {}

/// Immutable, but disallows `retain`
struct ImmutableWithMutableSubclass;
impl ClassKind for ImmutableWithMutableSubclass {}

struct Mutable;
impl ClassKind for Mutable {}

/// Mutable, but only on the main thread. `Id::retain` is safe
struct MainThreadOnly;
impl ClassKind for MainThreadOnly {}

trait ClassType {
    type Kind: ClassKind;
    // ... Super + NAME

    fn alloc()
    where Self::Kind != MainThreadOnly || Unknown;

    fn alloc(mtm: MainThreadMarker)
    where Self::Kind == MainThreadOnly;
}

// Maybe even:
unsafe impl Send + Sync for Immutable {}
unsafe impl Send + Sync for ImmutableWithMutableSubclass {}
unsafe impl Send + Sync for Mutable {}
```

And with a bit of extra trait bounds, we could even ensure that e.g. an `Immutable` class cannot subclass a `Mutable` one.

Still very unsure about this!

madsmtm:
With that, we could perhaps even get rid of the `Ownership` parameter on `Id`?
```rust
[...]

Issue #316, opened 2022-12-27
Improving mapping of generic types


madsmtm:
Currently, we're mapping generic types roughly as:

```rust
extern_class!(
    #[derive(PartialEq, Eq, Hash)]
    pub struct NSArray<T: Message = Object, O: Ownership = Shared> {
        p: PhantomData<Id<T, O>>,
    }

    unsafe impl<T: Message, O: Ownership> ClassType for NSArray<T, O> {
        type Super = NSObject;
    }
);

impl<T: Message, O: Ownership> NSArray<T, O> {
    fn get(&self, index: usize) -> Option<&T>;
}

impl<T: Message> NSArray<T, Owned> {
    fn get_mut(&mut self, index: usize) -> Option<&mut T>;
}
```

But this can be a bit limiting, since there's no easy way to carry any lifetime information in the type. Also, it is IMO a bit confusing to see `Id<NSArray<NSObject, Shared>, Shared>`, and this only gets worse on types with multiple generics (like `NSDictionary`).

So perhaps we could instead introduce a helper trait and do it like this:

```rust
trait Foo: Sealed {
    type Item: Message;
    type Ownership: Ownership;
}

impl<T: Message> Foo for &T {
    type Item = T;
    type Ownership = Shared;
}
impl<T: Message> Foo for &mut T {
    type Item = T;
    type Ownership = Owned;
}
impl<T: Message, O: Ownership> Foo for Id<T, Owned> {
    type Item = T;
    type Ownership = O;
}

// Usage

extern_class!(
    #[derive(PartialEq, Eq, Hash)]
    pub struct NSArray<T: Foo> {
        p: PhantomData<T>,
    }

    unsafe impl<T: Foo> ClassType for NSArray<T> {
        type Super = NSObject;
    }
);

impl<T: Foo> NSArray<T> {
    fn get(&self, index: usize) -> Option<&T>;
}

impl<T: Foo<Owneship = Owned>> NSArray<T> {
    fn get_mut(&mut self, index: usize) -> Option<&mut T>;
}
```

Where usage would then be `Id<NSArray<&NSObject>, Shared>` or `Id<NSArray<Id<NSObject, Shared>>, Shared>`.

This will also help with https://github.com/madsmtm/objc2/issues/304, since `-[NSEnumerator nextObject]` by itself could now be safe by default (though `-[NSArray objectEnumerator]` would still be `unsafe`).

madsmtm:
[...]

Issue #265, opened 2022-09-08
The deal with mutability


madsmtm:
I was [once convinced](https://github.com/SSheldon/rust-objc/issues/112#issuecomment-1139443814) that we could restrict mutation to `&mut self` methods, but after [having battled with AppKit](https://github.com/rust-windowing/winit/pull/2465) and seen how Swift does it, I'm beginning to doubt that this is the best way.

`NSString` is immutable while `NSMutableString` is, naturally, mutable. Since our current implementation only allows mutation of `NSMutableString` through `&mut self`, it is safe to make both of these `Send + Sync`; all is well and Rusty.

Now, remember that `&NSMutableString -> &NSString` and `Id<NSMutableString, O> -> Id<NSString, O>` is safe - so if we were to make mutation of `NSMutableString` possible through `&self`, not only would `NSMutableString` be `!Sync`, but so would `NSString`! (They could still be `Send`, that would simply allow moving them via. `Id<T, Owned>` to a different thread - quite similar to `std::cell::Cell`)

That's the primary downside: Our objects can no longer be shared across threads. Downsides that usually apply to Rust code (aliasing optimizations, ...) are void in our case, since `NSString` is already `UnsafeCell`.

On the other hand, if we were to remove mutability we could:
1. Make `NSArray<T>`, `NSDictionary<K, V>` and such collection types simpler
2. Since most frameworks use interior mutability and/or are not thread safe anyhow, the usage of the Foundation framework would match the usage of these.
3. It would be possible for users to inherit `NSString`, without them having to ensure that their object was `Sync`
4. `Id::retain` could be made safe(r?)
5. [automatic](https://github.com/madsmtm/objc2/pull/264) [bindings](https://github.com/madsmtm/objc2/issues/85) would be simpler
6. Using `NSMutableString` inside a `declare_class!` that is not meant to be thread-safe anyhow is easier (e.g. means we won't have to use `&mut` in `winit`)

So... Yeah, will think about this a bit, but I think we may have to sacrifice being able to use Objective-C classes across threads (exactly the same compromise Swift does, their `String` is `Sendable` but `NSString` is not).

madsmtm:
https://github.com/rust-windowing/winit/pull/2457 highlighted an issue: We'd have to return `Id<WinitView, Owned>` if we wanted to make `accepts_first_mouse` mutable after the fact - but you can very rarely have an owned view (because other code may hold references to it, unknown to you), so this would be a footgun.

[...]

PR #419, opened 2023-02-04
Redo how ownership works

madsmtm:
Add associated type `ClassType::Mutability`, which describes the mutability of an object. This replaces the `Ownership` parameter in `Id<T, O>` (so we're moving the mutability/ownership specification from use site to declaration site, which is a bit more restrictive, but much more succinct, as well as being easier to understand and harder to make mistakes).

Along with this, we can also finally fix `NSCopying` and `NSMutableCopying` in a way that `ProtocolObject<dyn NSCopying>` is allowed.

The rough idea was initially described in https://github.com/madsmtm/objc2/issues/399#issuecomment-1413666160, see the code and changelog for details on how the exact design turned out.

- Fixes https://github.com/madsmtm/objc2/issues/399 by creating `ClassType::retain`, which is safe for certain classes.
- Fixes https://github.com/madsmtm/objc2/issues/316 by removing the `Ownership` type on generics.
- Fixes https://github.com/madsmtm/objc2/issues/401 by removing the associated types from `NSCopying` and `NSMutableCopying`, and instead having that data encoded in the associated type `ClassType::Mutability`.
- Fixes https://github.com/madsmtm/objc2/issues/265 by disallowing `Id<T> -> &mut T` when `T` is not mutable. The remaining part in that (proper main thread safety) is tracked in https://github.com/madsmtm/objc2/pull/359.


madsmtm:
Which kind of subclassing should be allowed:
```rust
Unknown -> *
Immutable -> Immutable
ImmutableWithMutableSubclass -> Immutable + ImmutableWithMutableSubclass + Mutable
Mutable -> Mutable (+ ImmutableWithMutableSubclass)
InteriorMutable -> InteriorMutable (+ MainThreadOnly)
MainThreadOnly -> MainThreadOnly (+ InteriorMutable)
```

madsmtm:
I'm wondering how to assign things in `icrate`, and I actually think we don't really need `Unknown`, rather it should be called `Root`, and instead we should just default everything to `InteriorMutable`?

madsmtm:
Also, `InteriorMutable` and `Immutable` are currently equivalent? (apart from `Send + Sync`, but those should be implemented separately)

Though perhaps we can still use it in e.g. `NSDictionary` to only allow `Immutable` keys, thereby fixing https://github.com/madsmtm/objc2/issues/337?

madsmtm:
I considered that it may be favourable to allow changing the class kind from the default to `Mutable` later on, and would like to make that a non-breaking change, but since we'd still need to change the methods from `&self` to `&mut self`, there is not really much value in that (e.g. might be relevant for `NSFormatter`?)

Changing from the default to `Immutable` is safe though!

madsmtm:
Tagging @simlay, @silvanshade and @mwcampbell as you might want to be notified of this change: `Id` no longer has an ownership parameter `O`, instead the information about ownership/mutability is moved to an associated constant `ClassType::Mutability`, which should feel significantly simpler to use.

The changelog should contain migration instructions, but please let me know if you run into any issues!

The change is primarily done because people (including myself) keep wanting to use `&mut self` in `declare_class!`, which is almost always unsound when overriding AppKit classes or implementing delegates - instead one should use `&self` + interior mutability (instance variables wrapped in `Cell`).

PR #435, opened 2023-04-11
Support for doc comments in icrate generation

simlay:
I'm not sure there's an issue for this but saw the `TODO` and felt like doing it. My biggest complaint is that the generated doc strings have various type annotations meant for something like sphinx. It makes them just kinda gross.

Also, some of the comments don't work great with some of the macros.


crates/header-translator/src/comment.rs
```diff
@@ -0,0 +1,126 @@
+//! Utilities for manipulating C/C++ comments.
```
simlay, line 1: I copied this [file straight out of bindgen](https://github.com/rust-lang/rust-bindgen/blob/a9681090bf1d925b3da7444e75a64dfc3e46a827/bindgen/ir/comment.rs) and then modified it for some of the extra gross objective-c comments.

madsmtm (changes requested):
Oh, I remember trying this before I discovered the "-fretain-comments-from-system-headers" flag, and then never got back to it, but this is really nice!

> My biggest complaint is that the generated doc strings have various type annotations meant for something like sphinx. It makes them just kinda gross.

Yeah, I can see a few options for resolving that:
1. [`Entity::get_parsed_comment`](https://docs.rs/clang/2.0.0/clang/struct.Entity.html#method.get_parsed_comment), if it handles those tags? If not, maybe it's possible to turn on a `clang` feature flag to handle that?
2. Shell out to some sort of external `sphinx` parser.
3. Investigate how Xcode handles them normally, maybe use their implementation (if it's open-source, probably difficult otherwise)

But it's still better than nothing.

> Also, some of the comments don't work great with some of the macros.

Yeah, `typed_enum`, `typed_extensible_enum` and `extern_static` need a `$(#[$m:meta])*` in their definition, should be fairly simple (see `icrate/src/macros.rs`).

madsmtm:
I've somewhat finished this work in 581f0986b2f1d7c3ea4495493a2094945c5cb83a (the output still isn't great, but it's better than nothing), thanks for starting on it @simlay.

PR #439, opened 2023-04-15
Allow passing `Id`'s as arguments to `msg_send!`

madsmtm:
Allow `Id<T>` as a message send parameter that is just dropped after the message send, such that taking a `T: IsMutable` class as argument, which would usually be unsafe since we could easily modify the class afterwards, now can be safe.


madsmtm:
It is not immediately clear to the user whether this means `ns_consumed` or the current semantics (i.e. whether the message send consumes the retain count or not), so they might mistakenly think that they can use this for `ns_consumed`.

I think the desired behaviour, while it could still be autogenerated by `header-translator` for `Mutable` classes, it should ultimately be handled outside of `extern_class!`.

PR #454, opened 2023-06-03
Fix instance variable object encoding

madsmtm:
Instance variables include extended type information, we should find some way to represent this.

See: https://bou.io/ExtendedTypeInfoInObjC.html


madsmtm:
While fairly easy to parse, I'm yet unsure of how it should be represented in `objc2-encode`.

PR #460, opened 2023-06-15
Alternatives to `ProtocolObject`

madsmtm:
I suspect it may be possible for us to basically never need to use `ProtocolObject`, but instead rely on the compiler to do the work of figuring out which traits are valid for which objects.

In this PR I've experimented with `CoercedUnsized` for `Id`, to show that it can indeed work through that. If we also changed message sending such that `&dyn NSObjectProtocol` translates to `NonNull<AnyObject>`, then using protocol objects would look very, _very_ similar to normal Rust code (methods take `&dyn MyProtocol` and end up returning `Id<dyn MyProtocol>`).

One worry is that the vtable may be included in the final binary, that would have to be thoroughly tested beforehand.

### DIFF
--- a/crates/objc2/src/lib.rs
+++ b/crates/objc2/src/lib.rs
@@ -158,6 +158,7 @@
 )]
 #![cfg_attr(feature = "unstable-c-unwind", feature(c_unwind))]
 #![cfg_attr(feature = "unstable-docsrs", feature(doc_cfg, doc_auto_cfg))]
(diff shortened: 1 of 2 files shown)


madsmtm:
I've experimented a bit with this in bb1ae6ff2ad58377a3ed2fc4f0cbeb83cd558b15, and decided to track the rest of this PR in https://github.com/madsmtm/objc2/issues/686.

PR #467, opened 2023-06-19
Add the `CoreAudioTypes` framework

samuelsleight:
This PR adds support for the [`CoreAudioTypes` framework](https://developer.apple.com/documentation/coreaudiotypes?language=objc), as an initial, (somewhat) simple start to the road of _eventually_ adding `AVFoundation`.

As simple and small a library as it is (almost exclusively type/constant definitions and header-only, not even any linking), it did throw a couple of curveballs at me:

- I needed to support four character codes, as they were used in several of the enum definitions in here. I've opted to do this by pulling [this library](https://github.com/shurizzle/rust-four-char-code) in as a dependency of the header translator, and outputting the underlying `u32` into the generated code so it's completely silent to the `icrate` library itself. I'm not 100% convinced I've done this in the best way, so I'm very much open to discussion here!
- One of the exposed constans required a cast as part of it's definition, so I needed to add that as a fix manually. Feels like that _should_ be a solvable problem, but I'm not sure it actually is at this stage?

With regards to the generated code, [I appear to have added a large number of `deprecated` attributes](https://github.com/samuelsleight/icrate-generated/commit/bd5e14aadb3b6c5fe3167b3631f76a2846e81046) - I'm not sure if this is some issue with my SDK setup or whether the submodule just hasn't been updated since some relevant change! Happy to look into that if needed.

More than happy for discussions on any decisions I've made in here, given I've not added anything here before!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bd0c93bb4b0c6d9b77f4435b0ae98c24d17f1c45b2ff844c6151a07256ca923b"
 
 [[package]]
+name = "four-char-code"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c661315fd366b2a1f970df7b7cb1a28d2678d49ef4872f7dcc19b4a83150f20b"
+
+[[package]]
 name = "glob"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -146,6 +152,7 @@ dependencies = [
  "basic-toml",
  "clang",
  "clang-sys",
+ "four-char-code",
  "heck",
  "proc-macro2",
  "semver",
--- a/crates/header-translator/Cargo.toml
+++ b/crates/header-translator/Cargo.toml
@@ -20,3 +20,4 @@ proc-macro2 = "1.0.49"
 syn = { version = "2.0", features = ["parsing"] }
 heck = "0.4"
 semver = { version = "1.0", features = ["serde"] }
+four-char-code = "2.2.0"
--- a/crates/header-translator/framework-includes.h
+++ b/crates/header-translator/framework-includes.h
@@ -37,6 +37,8 @@
 
 #import <Contacts/Contacts.h>
 
+#import <CoreAudioTypes/CoreAudioTypes.h>
+
 #import <CoreData/CoreData.h>
 
 #import <CoreLocation/CoreLocation.h>
--- a/crates/header-translator/src/expr.rs
+++ b/crates/header-translator/src/expr.rs
@@ -132,6 +132,14 @@ impl Expr {
             s = s.trim_start_matches('(').trim_end_matches(')').to_string();
         }
 
+        // Handle four character codes
+        if s.len() == 6 && s.starts_with('\'') && s.ends_with('\'') {
+            let fcc = four_char_code::FourCharCode::from_str(&s[1..5])
+                .expect("Invalid four character code");
+
+            return Some(Self::Unsigned(fcc.as_u32().into()));
+        }
+
         Some(Self::String(s))
     }
 }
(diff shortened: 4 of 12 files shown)


madsmtm:
>     * I needed to support four character codes, as they were used in several of the enum definitions in here. I've opted to do this by pulling [this library](https://github.com/shurizzle/rust-four-char-code) in as a dependency of the header translator, and outputting the underlying `u32` into the generated code so it's completely silent to the `icrate` library itself. I'm not 100% convinced I've done this in the best way, so I'm very much open to discussion here!

I think it's best to put it in `header-translator`, the reasoning mostly being "we can, and it's faster for consumers". I'll definitely accept this code!

>     * One of the exposed constans required a cast as part of it's definition, so I needed to add that as a fix manually. Feels like that _should_ be a solvable problem, but I'm not sure it actually is at this stage?

The reason is that the anonymous enum ends up having a different type from the real enum. I think the general fix would just be to always cast when defining enum constants. I've just sorta assigned it as "part of https://github.com/madsmtm/objc2/issues/310" for now.

> With regards to the generated code, [I appear to have added a large number of `deprecated` attributes](https://github.com/samuelsleight/icrate-generated/commit/bd5e14aadb3b6c5fe3167b3631f76a2846e81046) - I'm not sure if this is some issue with my SDK setup or whether the submodule just hasn't been updated since some relevant change! Happy to look into that if needed.

Yeah, the bindings are currently using the SDK from Xcode 14.2. I can update it to Xcode 14.3.1, but that requires me to do a 7GB download on a metered connection (I'm outside my home country currently), so I've been reluctant to do so. I'll do it at the start of August in any case, but if it's bothering you a lot, I can do it now.

> More than happy for discussions on any decisions I've made in here, given I've not added anything here before!

You've done great so far, thank you!

samuelsleight:
> think the general fix would just be to always cast when defining enum constants.

I did actually contemplate this one, though when I tried it I ran into issues when we special-case `NSIntegerMax` and `NSUIntegerMax` - though having said that, it would be equally trivial to un-special case that! I'm happy to try that if you think it's a reasonable solution?

> Yeah, the bindings are currently using the SDK from Xcode 14.2

The odd thing is that in _theory_ I _am_ using the 14.2 SDK, as I noted on [my other PR](https://github.com/madsmtm/objc2/pull/468) - there I just didn't commit any of the `deprecated` additions, and that seems to have worked fine, so I cna do the same here and only commit the `CoreAudioTypes` addition which I expect would also work (which I did split into a separate commit for this purpose), but I _would_ like to attempt to work out why I'm seeing the difference.

samuelsleight:
> Ah, then I think the issue is that you're using a newer clang than I am, which converts API_DEPRECATED attributes on the `@interface` into deprecated attributes on the methods as well.

Ahhhhhh, that makes sense! If we kept it, would it then be annoying for those that _don't_ have the newer version?

madsmtm:
> would it then be annoying for those that don't have the newer version?

I suspect that's only me, but I guess we could also add a hack for doing the propagation ourselves?

samuelsleight:
> but I guess we could also add a hack for doing the propagation ourselves?

That could work, to be fair, though it depends how worth it it is likely to be versus how soon you'd be updating anyway, I guess

madsmtm:
Yeah. Though actually, I think our CI currently uses the clang from Xcode 14.2, so we'd probably have to either bump or support that too? So maybe it actually _does_ make sense to add this hack, if not only to set some kind of precedent for how easy we make it to contribute to `header-translator`?

samuelsleight:
Ah yeah, that's a good point, and I'm definitely in favour of making things easier for people if possible

samuelsleight:
@madsmtm cheers for the update of xcide etc - that should definitely make things easier!

Apologies, I'd not (intentionally) abandoned these PRs, I'd just been relatively busy recently - should be able to bring them at least up to date at some point this week though

madsmtm:
I've added the four character code support in 4961dda39e2590ffb70c470628f41a508ef7b93d and the `CoreAudioTypes` in c89bf6f6e7732d89417fa60db92941dec08b2e01, both based on your implementation, thanks for your work!

samuelsleight:
Wow, I really did say "this week" over a year ago huh

@madsmtm cheers for utilising some of this and https://github.com/madsmtm/objc2/pull/471 despite me definitely abandoning them, intentionally or not!

PR #471, opened 2023-06-24
Add the `CoreFoundation` framework

samuelsleight:
This PR is a recreation of https://github.com/madsmtm/objc2/pull/400, to add the `CoreFoundation` framework

The steps to produce this were to add the requisite boilerplate, make the changes needed for the generated code to compile, make changes on a per-framework basis for all our existing frameworks to compile, and then to finally fix any remaining errors in the header translator output.

Going commit-by-commit should break this down into those steps nicely, and hopefully there should be nothing particularly controversial about this, though there will likely be some discussion points about specfic details

Only other thing to note is that I still have the same issue I've previously been having with my generated code, so I've ~not yet pushed up a submodule change to this PR~ pushed up the submodule change but expect the CI comparison to fail - I do however plan to start [looking into resolving that](https://github.com/madsmtm/objc2/pull/467#issuecomment-1599438017) next


samuelsleight:
Definitely should have added the GNUStep target to my local testing!

@madsmtm Is there a way to require some generated code be behind a feature flag? I can easily enough go and add `#[cfg(feature = "apple")]` to the requisite `libc` imports, but that wouldn't fix the generated code that _uses_ those imports. Is there a way, or do I need to manually copy those into `fixes` and add them there?

madsmtm:
Just responding to GNUStep question, will look at the rest later.

I think you may be able to require the feature flag on functions/methods using a specific type by changing `visit_required_types`?

madsmtm:
I've rebased and modified this to become https://github.com/madsmtm/objc2/pull/676, thanks for your initial work on this!

PR #473, opened 2023-07-26
Update proc-macro dep so it compiles on nightly again

cynecx:
Not sure which particular proc-macro release fixed this but from their changelog I can see that some releases did have some changes around nightly features.

Edit: https://github.com/dtolnay/proc-macro2/releases/tag/1.0.60 probably fixed this.


madsmtm:
Thanks!

PR #474, opened 2023-07-26
Add downcasting capability based on `isKindOfClass:`

cynecx:
Add downcasting capability based on `isKindOfClass:`.

### DIFF
--- a/crates/objc2/src/runtime/nsobject.rs
+++ b/crates/objc2/src/runtime/nsobject.rs
@@ -115,8 +115,10 @@ pub unsafe trait NSObjectProtocol {
     /// Check if the object is an instance of the class, or one of its
     /// subclasses.
     ///
-    /// See [Apple's documentation][apple-doc] for more details on what you
-    /// may (and what you may not) do with this information.
+    /// See [`AnyObject::downcast`] if your intention is to use this to cast
+    /// an object to another, and see [Apple's documentation][apple-doc] for
+    /// more details on what you may (and what you may not) do with this
+    /// information in general.
     ///
     /// [apple-doc]: https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418511-iskindofclass?language=objc
     #[doc(alias = "isKindOfClass:")]
--- a/crates/objc2/src/runtime/nsobject.rs
+++ b/crates/objc2/src/runtime/nsobject.rs
@@ -115,10 +115,10 @@ pub unsafe trait NSObjectProtocol {
     /// Check if the object is an instance of the class, or one of its
     /// subclasses.
     ///
-    /// See [`AnyObject::downcast`] if your intention is to use this to cast
-    /// an object to another, and see [Apple's documentation][apple-doc] for
-    /// more details on what you may (and what you may not) do with this
-    /// information in general.
+    /// See [`AnyObject::downcast_ref`] or [`Retained::downcast`] if your
+    /// intention is to use this to cast an object to another, and see
+    /// [Apple's documentation][apple-doc] for more details on what you may
+    /// (and what you may not) do with this information in general.
     ///
     /// [apple-doc]: https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418511-iskindofclass?language=objc
     #[doc(alias = "isKindOfClass:")]
@@ -133,7 +133,7 @@ pub unsafe trait NSObjectProtocol {
     /// subclasses.
     ///
     /// See [`isKindOfClass`][Self::isKindOfClass] for details.
-    #[deprecated = "use `isKindOfClass` directly, or cast your objects with `AnyObject::downcast`"]
+    #[deprecated = "use `isKindOfClass` directly, or cast your objects with `AnyObject::downcast_ref`"]
     fn is_kind_of<T: ClassType>(&self) -> bool
     where
         Self: Sized + Message,
--- a/crates/objc2/src/runtime/nsobject.rs
+++ b/crates/objc2/src/runtime/nsobject.rs
@@ -236,7 +236,7 @@ pub unsafe trait NSObjectProtocol {
     ///
     /// # let obj = NSObject::new();
     /// // SAFETY: Descriptions are always `NSString`.
-    /// let desc: Retained<NSString> = unsafe { Retained::cast(obj.description()) };
+    /// let desc: Retained<NSString> = unsafe { Retained::cast_unchecked(obj.description()) };
     /// println!("{desc:?}");
     /// ```
     fn description(&self) -> Retained<NSObject>
@@ -467,7 +467,7 @@ mod tests {
 
     impl FakeSubclass {
         fn new() -> Retained<Self> {
-            unsafe { Retained::cast(NSObject::new()) }
+            Retained::downcast(NSObject::new()).unwrap()
         }
     }
 
--- a/crates/objc2/src/runtime/nsobject.rs
+++ b/crates/objc2/src/runtime/nsobject.rs
@@ -4,6 +4,7 @@ use core::hash;
 use crate::ffi::NSUInteger;
 use crate::rc::{Allocated, DefaultRetained, Retained};
 use crate::runtime::{AnyClass, AnyObject, AnyProtocol, ImplementedBy, ProtocolObject, Sel};
+use crate::DowncastTarget;
 use crate::{
     extern_methods, msg_send, msg_send_id, AllocAnyThread, ClassType, Message, ProtocolType,
 };
@@ -50,6 +51,8 @@ unsafe impl ClassType for NSObject {
         &self.__inner
     }
 }
+
+unsafe impl DowncastTarget for NSObject {}
 
 /// The methods that are fundamental to most Objective-C objects.
 ///
--- a/crates/objc2/src/runtime/nsobject.rs
+++ b/crates/objc2/src/runtime/nsobject.rs
@@ -131,9 +131,7 @@ pub unsafe trait NSObjectProtocol {
     /// subclasses.
     ///
     /// See [`isKindOfClass`][Self::isKindOfClass] for details.
-    #[doc(alias = "isKindOfClass")]
-    #[doc(alias = "isKindOfClass:")]
-    // TODO: Consider deprecating this
+    #[deprecated = "use `isKindOfClass` directly, or cast your objects with `AnyObject::downcast`"]
     fn is_kind_of<T: ClassType>(&self) -> bool
     where
         Self: Sized + Message,
@@ -536,12 +534,12 @@ mod tests {
     #[test]
     fn test_is_kind_of() {
         let obj = NSObject::new();
-        assert!(obj.is_kind_of::<NSObject>());
-        assert!(!obj.is_kind_of::<RcTestObject>());
+        assert!(obj.isKindOfClass(NSObject::class()));
+        assert!(!obj.isKindOfClass(RcTestObject::class()));
 
         let obj = RcTestObject::new();
-        assert!(obj.is_kind_of::<NSObject>());
-        assert!(obj.is_kind_of::<RcTestObject>());
+        assert!(obj.isKindOfClass(NSObject::class()));
+        assert!(obj.isKindOfClass(RcTestObject::class()));
     }
 
     #[test]
--- a/crates/objc2/src/runtime/nsobject.rs
+++ b/crates/objc2/src/runtime/nsobject.rs
@@ -141,15 +141,6 @@ pub unsafe trait NSObjectProtocol {
         self.isKindOfClass(T::class())
     }
 
-    // Note: We don't provide a method to convert `NSObject` to `T` based on
-    // `is_kind_of`, since that is not possible to do in general!
-    //
-    // For example, something may have a return type of `NSString`, while
-    // behind the scenes they really return `NSMutableString` and expect it to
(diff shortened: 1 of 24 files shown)


crates/objc2/src/runtime/nsobject.rs
```diff
@@ -155,6 +163,36 @@ pub unsafe trait NSObjectProtocol {
     // For example, something may have a return type of `NSString`, while
     // behind the scenes they really return `NSMutableString` and expect it to
     // not be modified.
```
madsmtm, line 165: Read the comment here; casting e.g. `Id<NSString>` to `Id<NSMutableString>`, even if that is the actual class, is not allowed!

Especially so because `Id<NSString>` is clonable, regardless of what the _actual_ inner class is! So using that, we could trivially violate memory safety:
```rust
let string: Id<NSMutableString> = NSMutableString::new();
let string: Id<NSString> = Id::into_super(string);
let clone: Id<NSString> = string.clone();

let string: Id<NSMutableString> = Id::member_cast(string);
let clone: Id<NSMutableString> = Id::member_cast(clone);

let string: &mut NSMutableString = &mut string;
let clone: &mut NSMutableString = &mut clone;
// `string` and `clone` are now mutable aliasing references, which is UB!
```

crates/objc2/src/runtime/nsobject.rs
```diff
@@ -155,6 +163,36 @@ pub unsafe trait NSObjectProtocol {
...
+
+    /// Checks if the object is a member of the class.
+    ///
+    /// See [Apple's documentation][apple-doc] for more details on what you
+    /// may (and what you may not) do with this information.
+    ///
+    /// [apple-doc]: https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418766-ismemberofclass
+    #[doc(alias = "isMemberOfClass")]
+    fn is_member_of<T: ClassType>(&self) -> bool
```
madsmtm, line 174: It seems to me like this is not _that_ useful? A faster implementation would basically just be `self.class() == T::class()`, which I also think is more understandable to a user than `is_member_of`/`isMemberOfClass`?

crates/objc2/src/runtime/nsobject.rs
```diff
@@ -155,6 +163,36 @@ pub unsafe trait NSObjectProtocol {
...
+        T: 'static,
+    {
+        if self.is_member_of::<T>() {
+            // SAFETY: self is a direct member of the specified (`T`) class.
+            // Therefore casting should be safe.
+            Some(unsafe { &*(self as *const Self).cast::<T>() })
+        } else {
+            None
+        }
+    }
```
madsmtm, line 195: Hmm, this one is _tricky_, and the reason I haven't added it yet is because I'm not 100% sure it would be sound! It seems like it _should_ be, since `&self` and `&T` are both immutable references, which avoids the main problem with `Id::member_cast`.

But in any case, it seems like it could just as safely use `__isKindOfClass` as `__isMemberOfClass`?

madsmtm (changes requested):

madsmtm:
Another danger is generic types! E.g. there are no safeguards to prevent you from getting `NSArray<NSString>` from e.g. `NSArray<NSData>`. The related restriction to `'static` types you've made so far is a good idea.

Also wanted to say that I agree that this feature is useful, and would like to get it _somehow_, it just needs a bit more thinking to make sure it's sound!

crates/objc2/src/runtime/nsobject.rs
cynecx: Ah, I see. The key information I was missing was that mutability is kinda "encoded" in class types. That's why I was under the impression that `Id` would be more like `Rc` (no `DerefMut`) but it's slightly different.

cynecx: Hm. But I feel like it should be possible to do this safely. Since we have knowledge of the complete "inheritance chain" and can only allow casts that don't skip over any `Mutable` classes 🤔.

crates/objc2/src/runtime/nsobject.rs
cynecx: I agree.

crates/objc2/src/runtime/nsobject.rs
cynecx: > But in any case, it seems like it could just as safely use __isKindOfClass as __isMemberOfClass?

I initially thought that having this restriction would reduce the common footgun described [here](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418511-iskindofclass#discussion). However, I also realize that our situation is slightly different since we have the concept of mutability here.

cynecx:
> Another danger is generic types!

😅 I have to admit I've started working with this crate since yesterday and didn't even realize that generics were a thing here.

There is one option where one could encode the "genericity" in `ClassType` and simply disallow such casts when the destination class contains any generics:

```rust
pub trait impl(self) Genericity {}

pub struct Generic(!);
impl Genericity for Generic {}

pub struct NotGeneric(!);
impl Genericity for NotGeneric {}

pub unsafe trait ClassType: Message {
    ...

    type Genericity: Genericity;
}

impl AnyObject {
    fn 🚲<T>(&self) -> Option<&T>
    where
        Self: 'static,
        T: ClassType<Genericity = NotGeneric> + 'static,
    {
        if self.class() == T::class() { // Or isKindOfClass
            // SAFETY: self is a direct member of the specified (`T`) class.
            // Therefore casting should be safe.
            Some(unsafe { &*(self as *const Self).cast::<T>() })
        } else {
            None
        }
    }
}
```

crates/objc2/src/runtime/nsobject.rs
madsmtm: No, the example they have:
```objective-c
// DO NOT DO THIS!
if ([myArray isKindOfClass:[NSMutableArray class]])
{
    // Modify the object
}
```

Is very, very similar to the problem that we have, and `isMemberOfClass:` suffers from it as well, e.g. the following is just as illegal (though likely just even worse, since `NSMutableArray` is in most runtimes a class cluster).
```objective-c
// DO NOT DO THIS!
if ([myArray isMemberOfClass:[NSMutableArray class]])
{
    // Modify the object
}
```

madsmtm: In general I would really like to steer users away from `isMemberOfClass:`, since it depends on what is arguably implementation details (the exact class the type ends up as), and towards `isKindOfClass:`, which is more robust to that.

cynecx: > No, the example they have:

Well, I think we agree on that part? That's why I said "I initially thought". The problem is still there, but the proposed api should prevent "modification" through immutability no?

madsmtm:
> one could encode the "genericity" in `ClassType`

Hmm... Yeah, that would go a long way in ensuring things are sound!

Although this does exclude us from doing `NSObject -> NSArray<AnyObject>`, which would be nice to support, so perhaps another approach would be better? Maybe we could add a new trait `BikeshedDowncastAble` which is implemented automatically in `extern_class!` and `declare_class!`, but for generic types we only implement it with the generic parameter being `AnyObject`?

Also, I'm considering whether it would be possible to not restrict this to `NSObjectProtocol`, such that you could also use it on `AnyObject`? Would be prudent to research what Swift does under the hood for [their type casting support](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/typecasting/).

madsmtm:
Something that would help my confidence in the soundness of whatever implementation we end up with, maybe we can assemble a list of allowed and disallowed cases (that could later on be used as UI tests)?

cynecx:
> Also, I'm considering whether it would be possible to not restrict this to NSObjectProtocol, such that you could also use it on AnyObject?

I think it might be possible. The objc runtime exposes an internal api called `objc_opt_isKindOfClass`:

- https://github.com/apple-oss-distributions/objc4/blob/c3f002513d195ef564f3c7e9496c2606360e144a/runtime/objc-internal.h#L1074
- https://github.com/apple-oss-distributions/objc4/blob/c3f002513d195ef564f3c7e9496c2606360e144a/runtime/NSObject.mm#L2104

Digging further it seems like NSObject's isKindOfClass implementation is also very similar to `objc_opt_isKindOfClass` one.

However, I am not quite sure about "how internal" that api is since it doesn't have an underscore like many other internal but yet exposed objc runtime apis. But it seems like llvm does an optimization where it transforms `[obj isKindOfClass]` calls to `objc_opt_isKindOfClass` calls (https://lldb.llvm.org/cpp_reference/AppleObjCTrampolineHandler_8cpp_source.html). Perhaps that's enough for us to consider using this. I'd rather avoid reimplementing internal implementation details here...

madsmtm:
I think `objc_opt_isKindOfClass` is just an optimization, and yeah, I am not inclined to bump the deployment target requirement just yet.

The problem with calling `isKindOfClass:` on `AnyObject` though is that if the object does _not_ subclass `NSObject`, then the method may not be available! But maybe we can just re-implement the loop ourselves in Rust?

I tried checking what Swift does, it seems like an expression like `obj as? NSString` (where `obj: AnyObject`) uses the runtime function [`swift_dynamicCastObjCClass`](https://github.com/apple/swift/blob/2eade1d1b553a645e475912167162e1f6dc5f78c/include/swift/Runtime/Casting.h#L77-L86), which [doesn't actually check whether `isKindOfClass:` is implemented](https://github.com/apple/swift/blob/2eade1d1b553a645e475912167162e1f6dc5f78c/stdlib/public/runtime/SwiftObject.mm#L1096). So maybe we don't really need to bother, since custom root classes are expected to be so rare anyhow? Unsure!

crates/objc2/src/runtime/nsobject.rs
cynecx: I omitted this in the latest push. You can already kinda do this with `obj.downcast::<...>().unwrap().retain()`.

cynecx:
Converted to a draft because it's probably still missing something (among others, ui-tests might be worthwhile).

madsmtm:
Okay so I tried the following Swift code:
```swift
import Foundation

let testConversionArray: AnyObject = NSArray.init(objects: NSString(), NSObject())

if let inner = testConversionArray as? [NSString] {
    print("is NSString array", inner[0].length)
} else if let inner = testConversionArray as? [NSObject] {
    print("is NSObject array", inner[0])
} else {
    print("not array")
}
```

And it is able to detect that the array _is_ actually a `NSObject` and not an `NSString` array.

The way that works is by [checking every element of the array](https://github.com/apple/swift/blob/main/docs/DynamicCasting.md#arraysetdictionary-casts) though, which is probably unacceptable performance-wise in Rust (I'd expect a downcast to be very cheap).

So in Rust users will probably have to do the cast more explicitly:
```rust
let array: &AnyObject = ...;
let array: &NSArray<AnyObject> = array.downcast()?;
for item in array {
    let item: &NSString = item.downcast()?;
    // Use item here.
}
```

Tl;dr: I still think we should disallow casts on generic items, except for when the generics are set to `AnyObject`.

madsmtm:
Returning back to this:

Decision: We shouldn't really care about `AnyObject` not being guaranteed to support `isKindOfClass:`. If feasible, we can try to insert extra checks (as is done in current impl), perhaps only with `debug_assertions` enabled, but it's not like it's unsound if we don't, the runtime is just gonna throw an exception (actually done by `CoreFoundation`, if that is not linked then we'll get an abort).

Decision: I think the name "downcast" makes sense, that is also [Swift's terminology](https://github.com/apple/swift/blob/main/docs/DynamicCasting.md), and matches the equivalent Rust concept of [`<Box<dyn Any>>::downcast`](https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.downcast-2).

Decision: We also want some sort of `Id::downcast<U: DowncastTarget>(this: Id<T>) -> Result<Id<U>, Id<T>>`, though unsure of the exact bound here yet.

Unsure how downcasting from `ProtocolObject` should work? `ProtocolObject::downcast`? Or should `downcast` be a method on `Message`? Can we downcast _to_ a `ProtocolObject` (probably doable with `conformsToProtocol:`), or should that be handled by something else?

Regarding mutability: An explicit design decision in `objc2` is that once you type-erase a mutable object, you are allowed to do normal reference-counting with it (that's also an implicit assumption in Objective-C, e.g. classes are allowed to pass `NSString` in their public API, while actually giving access to a `NSMutableString`). I feel this still needs more thought (and tests) before I'm confident that the way we do this is sound.

Also unsure if we could add some kind of `downcast_mut`? Where could it be used? Is `&mut NSString` -> `&mut NSMutableString` fine? Or is it only possible for mutable to mutable, e.g. `&mut NSMutableData` -> `&mut NSPurgeableData`?

I also still have to consider generics some more. Likely we'll be implementing the trait for `NSArray<AnyObject>`, but that still makes it hard to access e.g. a type-erased `NSMutableArray<NSMutableString>` - But then again, that's probably how it _should_ be, as after that object is type-erased, there is no way to know if it's not shared elsewhere.

---

I'll try to make a collection of examples and UI tests in one of the coming days to move forwards here.

madsmtm:
I've _finally_ been able to come back to this properly, I've waited so long because I wanted to [remove mutability first](https://github.com/madsmtm/objc2/issues/563), I felt it was the only real way I could feasibly ensure that this was sound.

madsmtm (approved):

PR #478, opened 2023-07-31
Allow `c_void` pointers as return and argument types in more cases (behind a feature flag)

madsmtm:
Will help with https://github.com/servo/core-foundation-rs/pull/628, since a lot of CoreFoundation types are used there as just `*mut c_void`, and it would be really cumbersome and error-prone to have to change all of these to custom type aliases with a proper `Encode` implementation.


madsmtm:
I'm considering implementing the logic in `objc2` instead, so that it doesn't pollute the "pure" `objc2-encode`, and it would allow us to change it more freely in later breaking releases

PR #484, opened 2023-08-22
Fix browser example

silvanshade:
It doesn't look like the browser example is currently being tested in CI either but I wasn't sure how to (re-)enable that.


madsmtm:
Yeah it's not tested in CI currently, I think that's because of:

https://github.com/madsmtm/objc2/blob/3d9b6864af05592fc18b1ab27b3da9ed3896bc53/.github/workflows/ci.yml#L98-L103

Fix looks good though

Issue #258, opened 2022-08-31
Make `ns_string!` macro work inside closures


madsmtm:
They don't currently for some reason!

Discovered in https://github.com/madsmtm/objc2/pull/254.

madsmtm:
Similar error when using:
```console
RUSTFLAGS="-Clink-dead-code" cargo build --example basic_usage
```

madsmtm:
I've cfg-gated the functionality behind the feature flag `unstable-static-nsstring` for now: https://github.com/madsmtm/objc2/pull/350

madsmtm:
[`objrs` adds special `export_name`s](https://gitlab.com/objrs/objrs/-/blob/b4f6598696b3fa622e6fddce7aff281770b0a8c2/frameworks/foundation/macros/src/lib.rs), so perhaps that's required?

madsmtm:
A change in `rustc` dead code elimination causes the `UTF16` to appear in the assembly output, even though it's known statically that the symbol is unused.

Note that this happens even if the `#[link_section = "..."]` attribute is removed.

Will probably need to bisect `rustc` to find the regression, it's somewhere between `nightly-2023-05-25` and `nightly-2023-08-27`.

madsmtm:
I bisected it to https://github.com/rust-lang/rust/pull/115126, probably https://github.com/rust-lang/rust/pull/115114, but also found that it only appears if you place the string in another static, not if you just use it as an expression.

So probably not worth doing more about.

complexspaces:
Sorry to necropost, but given the Rust bisection work you did last year what is the status of the `unstable-static-nsstring` flag? Is this something that can be tested as correct now?

madsmtm:
I doubt it, I suspect the two statics that's needed for this, the one in `__DATA,__cfstring` and the one in `__TEXT,__cstring,cstring_literals`, still need to end up in the same codegen unit for the linker to be able to pick this up for sure, and I'm not sure how to make Rust guarantee that happens?

PR #487, opened 2023-08-27
Update to Xcode 14.3, and update rustc nightly

madsmtm:
A change in `rustc` dead code elimination causes the `UTF16` in `ns_string!` to appear in the assembly output; tracking that in https://github.com/madsmtm/objc2/issues/258.

UI tests triage: Have mostly improved, the changed ordering for "help: the following other types implement trait ..." are nice, except for `MethodImplementation`, that one is still mostly gibberish.
Also, spans inside `declare_class!` have now been fixed, such that the erring function is reported, which is _really_ nice!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,6 +222,9 @@ jobs:
     - fmt
     - lint
 
+    env:
+      TEST_OVERWRITE: 1
+
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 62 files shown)


madsmtm:
@simlay, @samuelsleight: I've updated the Xcode version to 14.3.1, along with running it with a newer `clang`, so it should be easier for you to contribute now.

Issue #483, opened 2023-08-21
[icrate] Help with the creation of a spinning triangle example


PicoLeaf:
I was trying to use icrate for metal context creation, but the complete lack of any guidance on the topic was rough especially since it isn't something I've done before (not on macos), so I'd be interested in contributing an example for this, I'm making this issue to get help.

The goal is a spinning triangle against a black background with each vertex assigned a different color.
The "spinning triangle part" would be done in the vertex shader instead of the main thread.
From what I gathered, there needs to be done a few things for this to work:

- Create a window (that's not too bad)
- Get the default metal device (that's also easy)
- Create a MTKView within the window (I havent figured that one out)
- load the shaders
- prepare the pipeline state
- setup a triangle buffer, aswell as a data buffer for communicating the timestep for the spinning
- enter the main event loop, this require creating a delegate of the MTKView if I'm not mistaken
- Handle all of the drawable logic and the buffer changes for the timestep

PLEASE tell me everything I've gotten wrong, I have very little clue of what I'm doing, I don't even know if all of this is possible with the current version of icrate!

silvanshade:
@PicoLeaf A metal example with some animation would be really interesting to have!

I think it should be possible to implement this with icrate in its current state without too much trouble.

I don't really know much about Metal but I was able to make a very minimal template which you should be able to use to create something more advanced with animation.

I uploaded it at my local fork under the [metal-example](https://github.com/silvanshade/objc2/tree/metal-example) branch under [examples/metal.rs](https://github.com/silvanshade/objc2/blob/metal-example/crates/icrate/examples/metal.rs)

You can run the example with the following command from the `objc2` project root:
```
cargo run --package=icrate --example=metal --features=unstable-example-metal
```

This should create a window with a static triangle that looks like this:
<img width="624" alt="Screenshot 2023-08-22 at 3 37 32 PM" src="https://github.com/madsmtm/objc2/assets/11022302/d65b4525-5635-4b73-b623-ee08b826d150">

There's no logic for creating an app menu or quitting through the UI without `ctrl-c` but you can look at the `browser.rs` example for how to implement that.

Hope that helps.

PicoLeaf:
Woah, this is awesome, it sure helps, thank you so much!

silvanshade:
By the way, one small thing I missed: `currentDrawable` is nullable, so it should return an `Option`. I've updated the example with the fix.

madsmtm:
[...]

PR #489, opened 2023-08-27
Metal example

silvanshade:
This PR adds the Metal API example corresponding to the discussion at https://github.com/madsmtm/objc2/issues/483.

@PicoLeaf I merged your commits and did some final refactoring. Thanks for adding the animation part in there and creating the other PR.

### DIFF
--- a/crates/header-translator/src/output.rs
+++ b/crates/header-translator/src/output.rs
@@ -39,6 +39,7 @@ impl Output {
         let mut macos_10_13_features: BTreeSet<String> = vec![
             "unstable-frameworks-macos-10-7".into(),
             "unstable-example-delegate".into(),
+            "unstable-example-metal".into(),
             "unstable-example-nspasteboard".into(),
             "unstable-example-speech_synthesis".into(),
         ]
--- a/crates/icrate/Cargo.toml
+++ b/crates/icrate/Cargo.toml
@@ -78,6 +78,12 @@ required-features = [
     "unstable-example-browser"
 ]
 
+[[example]]
+name = "metal"
+required-features = [
+    "unstable-example-metal"
+]
+
 [features]
 default = ["std", "apple"]
 
@@ -162,6 +168,25 @@ unstable-example-browser = [
     "WebKit",
     "WebKit_WKNavigation",
     "WebKit_WKWebView",
+]
+unstable-example-metal = [
+    "apple",
+    "AppKit",
+    "AppKit_NSWindow",
+    "Foundation",
+    "Foundation_NSCoder",
+    "Foundation_NSError",
+    "Foundation_NSNotification",
+    "Foundation_NSString",
+    "Metal",
+    "Metal_MTLCompileOptions",
+    "Metal_MTLRenderPassDescriptor",
+    "Metal_MTLRenderPipelineColorAttachmentDescriptor",
+    "Metal_MTLRenderPipelineColorAttachmentDescriptorArray",
+    "Metal_MTLRenderPipelineDescriptor",
+    "MetalKit",
+    "MetalKit_MTKView",
+    "Metal_MTLRenderPassDescriptor",
 ]
 
 # Helps with CI
--- a/crates/icrate/Cargo.toml
+++ b/crates/icrate/Cargo.toml
@@ -178,6 +178,7 @@ unstable-example-metal = [
     "Foundation_NSError",
     "Foundation_NSNotification",
     "Foundation_NSString",
+    "Foundation_NSDate",
     "Metal",
     "Metal_MTLCompileOptions",
     "Metal_MTLRenderPassDescriptor",
--- a/crates/icrate/Cargo.toml
+++ b/crates/icrate/Cargo.toml
@@ -5311,6 +5311,7 @@ unstable-frameworks-macos-10-13 = [
     "Metal_all",
     "PhotoKit_all",
     "unstable-example-delegate",
+    "unstable-example-metal",
     "unstable-example-nspasteboard",
     "unstable-example-speech_synthesis",
     "unstable-frameworks-macos-10-7",
--- a/crates/icrate/examples/metal.rs
+++ b/crates/icrate/examples/metal.rs
@@ -9,11 +9,14 @@ use icrate::{
         NSBackingStoreBuffered, NSWindow, NSWindowStyleMaskClosable, NSWindowStyleMaskResizable,
         NSWindowStyleMaskTitled,
     },
-    Foundation::{NSNotification, NSObject, NSObjectProtocol, NSPoint, NSRect, NSSize, NSString},
+    Foundation::{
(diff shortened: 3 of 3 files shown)


crates/icrate/Cargo.toml
```diff
@@ -163,6 +169,26 @@ unstable-example-browser = [
     "WebKit_WKNavigation",
     "WebKit_WKWebView",
 ]
+unstable-example-metal = [
```
madsmtm, line 172: If you add this to `header-translator/src/output.rs` under e.g. `macos_11_features` (or whatever version is required), then the CI will be check the example

crates/icrate/examples/metal.rs
```diff
@@ -0,0 +1,344 @@
...
+                unsafe { Id::retain(ptr) }.expect("Failed to get default system device.")
+            };
+
+            // create the command queue
+            let command_queue = device
+                .newCommandQueue()
+                .expect("Failed to create a command queue.");
+
+            // create the app window
+            let window = {
```
madsmtm, line 117: This is sadly usually the wrong way to do it; windows should be initialized in or after `(applicationDidFinishLaunching:`, and unsure about this, but the same probably goes for a lot of the other metal stuff.

At the very least, a comment about this would be nice.

crates/icrate/examples/metal.rs
```diff
@@ -0,0 +1,344 @@
...
+        unsafe { msg_send_id![Self::alloc(), initWithShaders: shaders] }
+    }
+}
+
+fn main() {
+    let app = unsafe { NSApplication::sharedApplication() };
+    unsafe { app.setActivationPolicy(NSApplicationActivationPolicyRegular) };
+
+    #[rustfmt::skip]
+    let shaders = ns_string!(r#"
```
madsmtm, line 290: Nit: Move the shader to a constant at the top of the file, so that it's easier to see what the Rust structs we define are used for.

crates/icrate/examples/metal.rs
```diff
@@ -0,0 +1,344 @@
...
+
+#[repr(C)]
+pub struct Position {
+    pub x: f32,
+    pub y: f32,
+    pub z: f32,
+}
+
+#[repr(C)]
+pub struct Color {
```
madsmtm, line 50: Please note that this has the same ABI as `MTLPackedFloat3` (also goes for `Position`).

crates/icrate/examples/metal.rs
```diff
@@ -0,0 +1,344 @@
...
+        #[allow(non_snake_case)]
+        unsafe fn mtkView_drawableSizeWillChange(&self, _view: &MTKView, _size: NSSize) {
+            // println!("mtkView_drawableSizeWillChange");
+        }
+    }
+);
+
+unsafe impl NSObjectProtocol for Delegate {}
+
+#[cfg(target_os = "macos")]
```
madsmtm, line 278: I think the whole example might be macOS only?

Though maybe we could make it work on iOS as well, by implementing `UIApplicationDelegate` too, and swapping out the `window` type?

crates/icrate/examples/metal.rs
silvanshade: Oh, I see, I missed those in the definitions. But do you want me to just make a comment about that or actually use them somehow? Since they don't seem to have any public fields or conversion traits, I don't see a safe way to use those directly in this case (i.e., wouldn't we need to do some transmuting or accessing elements at memory offsets unsafely?)

madsmtm: > Since they don't seem to have any public fields or conversion traits

I think that's something to fix, but for now just commenting that they're the same is enough.

crates/icrate/examples/metal.rs
silvanshade: Yeah, I wasn't sure about this either.

One thing that might be a problem though is how do I actually modify the ivars in that case since `applicationDidFinishLaunching` takes `&self`?

I saw the comments for the recent changes about interior mutability and using `Cell` or `RefCell` but when I tried an example earlier I couldn't figure out how to get anything with that to actually work.

Can you maybe give an example for how I might be able to do that? (What would need to change with the struct definition, etc.)

madsmtm: I think you'd have to do `window: IvarDrop<Box<RefCell<Option<Id<NSWindow>>>>, "_window">`, however ugly that is :/

I'm working on getting rid of the `Box`, but the `RefCell` and the `Option` are basically unavoidable due to how Rust ownership works. I might consider making a helper tool for this very common case of "initialize later", but maybe I'll also just defer it to a higher-level crate like [`cacao`](https://github.com/ryanmcgrath/cacao).

silvanshade:
Okay, I think I've addressed all the comments.

crates/icrate/examples/metal.rs
```diff
@@ -0,0 +1,365 @@
...
+            let Some(pass_descriptor) = (unsafe { mtk_view.currentRenderPassDescriptor() }) else { return; };
+            let Some(encoder) = command_buffer.renderCommandEncoderWithDescriptor(&pass_descriptor) else { return; };
+
+            // compute the scene properties
+            let mut scene_properties_data = SceneProperties {
+                time: unsafe { self.start_date.timeIntervalSinceNow() } as f32,
+            };
+            // write the scene properties to the vertex shader argument buffer at index 0
+            let scene_properties_bytes =
+                unsafe { NonNull::new_unchecked(&mut scene_properties_data) };
```
madsmtm, line 263: Hmm, I think these are copied by `setVertexBytes_length_atIndex`. So probably would be better to use `&` and the [`From<&T>` impl on `NonNull`](https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html#impl-From%3C%26T%3E-for-NonNull%3CT%3E) here instead.

```suggestion
            let scene_properties_data = SceneProperties {
                time: unsafe { self.start_date.timeIntervalSinceNow() } as f32,
            };
            // write the scene properties to the vertex shader argument buffer at index 0
            let scene_properties_bytes = NonNull::from(&scene_properties_data);
```

Actually, maybe we should derive `Copy` for `SceneProperties`, just to make that copying very clear?

Same goes for `VertexInput` stuff below.

madsmtm (commented):
Nice, just one more, then I _think_ we're good

crates/header-translator/src/output.rs
```diff
@@ -39,6 +39,7 @@ impl Output {
         let mut macos_10_13_features: BTreeSet<String> = vec![
             "unstable-frameworks-macos-10-7".into(),
             "unstable-example-delegate".into(),
+            "unstable-example-metal".into(),
```
madsmtm, line 42: To make CI pass, I think you need to add this to `icrate/Cargo.toml` as well (as the diff suggests).

madsmtm (approved):
Perfect, will merge once CI passes

PR #490, opened 2023-09-01
More refactoring for examples

silvanshade:
This is just some small fixes for the `metal` example and then a rewrite of the `browser` example to use deferred initialization via `RefCell` like with the `metal` example.


madsmtm (approved):
Nice, thanks!

PR #498, opened 2023-09-03
Refactor browser and metal examples to use OnceCell

silvanshade:
This PR refactors the `browser` and `metal` examples to use `OnceCell` instead of `RefCell`.

I feel this is a more precise representation of how those ivars are used since they are intended to only be initialized once, and it should be an error to try and set them again (in this case, although in other scenarios `RefCell` could certainly make sense).

As a nice side effect, it should be a little more efficient and we don't seem to need the inner `Option` either.

### DIFF
--- a/crates/icrate/examples/browser.rs
+++ b/crates/icrate/examples/browser.rs
@@ -1,5 +1,5 @@
 #![deny(unsafe_op_in_unsafe_fn)]
-use core::{cell::RefCell, ptr::NonNull};
+use core::{cell::OnceCell, ptr::NonNull};
 
 use icrate::{
     AppKit::{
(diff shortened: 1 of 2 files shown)


crates/icrate/examples/browser.rs
```diff
@@ -26,20 +26,30 @@ use objc2::{
...
 };
 
-type IdCell<T> = Box<RefCell<Option<Id<T>>>>;
+type IdCell<T> = Box<OnceCell<Id<T>>>;
 
 macro_rules! idcell {
+    ($name:ident => $this:expr) => {
+        $this.$name.set($name).expect(&format!(
+            "ivar should not already be initialized: `{}`",
+            concat!(stringify!($name))
```
madsmtm, line 35: Nit, `concat` not required. Same below.

```suggestion
            stringify!($name)
```

crates/icrate/examples/browser.rs
```diff
@@ -26,20 +26,30 @@ use objc2::{
...
+        ));
+    };
     ($name:ident <= $this:expr) => {
-        let $name = $this.$name.borrow();
-        let $name = $name
-            .as_ref()
-            .expect(concat!(stringify!($name), " ivar should be initialized"));
+        #[rustfmt::skip]
+        let Some($name) = $this.$name.get() else {
+            // NOTE: could use `core::hint::unreachable_unchecked` here
```
madsmtm, line 41: Well, yes, but only if the `idcell!` macro was `unsafe` to use. So I think it's fine to keep as `unreachable!`.

madsmtm (approved):
Ohh nice, since it's so newly stabilized I hadn't really internalized that `OnceCell` exist

PR #519, opened 2023-09-24
Don't insert receiver NULL checks on GNUStep

madsmtm:


madsmtm:
Turns out this was done in f4f95a447b1077686ecd1e09f01c952ca8d9411e for a reason, I'll document it more properly in a follow-up PR

Issue #455, opened 2023-06-09
Calling super methods safely


madsmtm:
Calling a super method is (with the below restrictions) as safe as calling regular methods.

Since doing so is a common requirement, e.g. in initializers, or when subclassing e.g. `NSView`, we should find a way so that the user can easily do so.

Requirements that I know of so far:
- Only safe when declaring a new subclass.
- Only direct super methods (e.g. calling `NSResponder`'s methods from a subclass of `NSView` could potentially lead to UB, if `NSView` was implemented in a certain way.
- Super method calls can only access methods on the superclass, not on the declared class.
- Initializer methods only from inside other initializer methods.

The main problem with this is that it effectively doubles the amount of methods that we have to generate, the codegen is different for each. Though maybe with https://github.com/madsmtm/objc2/issues/448, that will be less of a problem?

If we had inherent traits, or something like it, then it would perhaps be easier, since we could make the implementation on a trait instead.

Related: Internally, we can consider getting rid of `send_message_super`, by providing a `struct Super<T: MessageReceiver> { obj: T, cls: &'static Class }` that implements `MessageReceiver::send_message` by calling `objc_msgSendSuper` instead.

Related:
- https://github.com/madsmtm/objc2/pull/173
- https://github.com/madsmtm/objc2/issues/438

madsmtm:
One last requirement that actually ends up making this work:
- Only need to call super methods when overriding a method (otherwise you'd just call the method normally)

Which means that we can do the super class method generation inside `declare_class!` instead of `extern_class!`!

madsmtm:
It would probably also be nice to require the user to add an `#[override]` attribute (matches Swift's required `override` keyword.

madsmtm:
Actually _doing_ the super method generation is super hard (haha).

E.g. for the following to work:
```rust
define_class!(
    #[unsafe(super(NSObject))]
    struct Foo;

    impl Foo {
        #[override]
        #[method(init)]
        fn init(this: Allocated<Self>) -> Retained<Self> {
            let this: PartialInit<Self> = this.set_ivars(());
            this.into_super().init()
        }

        #[override]
        #[method(description)]
        fn description(&self) -> Retained<NSString> {
            self.as_super().description()
        }
    }
);
```

We'd need to output something like
```rust
#[repr(transparent)]
struct SuperInitHelper(PartialInit<Foo>);

#[repr(transparent)]
struct SuperHelper(Foo);

struct init;
[...]

Issue #438, opened 2023-04-13
`declare_class!`: Safer initializers


madsmtm:
Use something similar to [Swift's rules for initializers](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/#Class-Inheritance-and-Initialization), which we may be able to with a combination of types and a `proc-macro`, to verify that initializers only access the things that they are allowed to.

Should also somehow make ivars safe to declare, by statically ensuring that they are initialized - i.e. this issue ties in heavily with https://github.com/madsmtm/objc2/issues/414).

See also the Objective-C [docs on initialization](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Initialization/Initialization.html#//apple_ref/doc/uid/TP40010810-CH6) and [a bit more about designated initializers](https://developer.apple.com/library/archive/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html#//apple_ref/doc/uid/TP40014150-CH1-SW8).

madsmtm:
Interestingly, Objective-C recommends setting custom instance variables _after_ calling the superclass' initializer, while Swift requires you to set them _before_ (as part of their nice two-phase initialization scheme).

I can think of ways to express either in Rust:
```rust
// Types in this example put after `let` statements for clarity; they would be inferred in real life.
//
// Note also the `?` after the `msg_send_id`, which is possible since the initializer always (?) returns an Option

declare_class!(
    struct MyClass {
        ivar1: IvarEncode<i32>,
        ivar2: IvarDrop<Id<NSString>>,
    }

    // Now also generates:
    // struct MyClassIvars {
    //     ivar1: i32,
    //     ivar2: Id<NSString>,
    // }
    //
    // struct MyClassPartialInit;

    unsafe impl ClassType for MyClass {
        type Super = NSObject;
        type Mutability = InteriorMutable;
        const NAME: &'static str ="MyClass";
    }

    unsafe impl MyClass {
        // Objective-C style
        #[method_id(init:withArg:)]
        fn init(this: Allocated<Self>, arg: i32) -> Option<Id<Self>> {
            let this: MyClassPartialInit = unsafe { msg_send_id![super(this), init] }?;
            let this: Id<Self> = this.set(MyClassIvars {
                ivar1: arg,
                ivar2: NSString::new(),
            });
            Some(this)
        }

        // Swift style
        #[method_id(init:withArg:)]
[...]

PR #520, opened 2023-09-28
Change `Allocated<T>` in preparation for arbitrary self types

madsmtm:
Arbitary self types (tracking issue: https://github.com/rust-lang/rust/issues/44874) are not yet stable, but whatever form they end up as, we'll need to use a different type than `Option<Allocated<T>>`, since you can't (and won't be able to) use that as `self: Option<Allocated<Self>>`.

~My proposal here is to introduce a new type, since we likely still need to be able to differentiate between "this type is newly allocated, and we don't know the nullability" (such as when sending to `init` methods) vs. "this type is allocated, and we know that it is non-null" (such as when receiving from `init` methods).~

EDIT: See [below](https://github.com/madsmtm/objc2/pull/520#issuecomment-1738839859) for the alternative I'm going with.

### DIFF
--- a/crates/header-translator/src/method.rs
+++ b/crates/header-translator/src/method.rs
@@ -686,7 +686,7 @@ impl fmt::Display for Method {
 
         // Receiver
         if let MemoryManagement::IdInit = self.memory_management {
-            write!(f, "this: Option<Allocated<Self>>, ")?;
+            write!(f, "this: Allocated<Self>, ")?;
         } else if self.is_class {
             // Insert nothing; a class method is assumed
         } else if self.mutating {
--- a/crates/icrate/src/additions/Foundation/data.rs
+++ b/crates/icrate/src/additions/Foundation/data.rs
@@ -267,7 +267,7 @@ impl IdFromIterator<u8> for NSMutableData {
 }
 
 #[cfg(feature = "block")]
-unsafe fn with_vec<T: Message>(obj: Option<Allocated<T>>, bytes: Vec<u8>) -> Id<T> {
+unsafe fn with_vec<T: Message>(obj: Allocated<T>, bytes: Vec<u8>) -> Id<T> {
     use core::mem::ManuallyDrop;
 
     use block2::{Block, ConcreteBlock};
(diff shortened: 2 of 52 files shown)


madsmtm:
I _think_ there might be an alternative path:

We use `struct Allocated<T>(Option<NonNull<T>>)` everywhere, and then inside `declare_class!` when [initializing instance variables](https://github.com/madsmtm/objc2/issues/438), we do a NULL check to ensure that the allocation succeeded (e.g. we're not out of memory).

With a bit of macro trickery, inlining and LLVM optimizations, we can probably fairly reliably make it so that this NULL check is omitted when the declared method is called from Objective-C (since we know that receivers can't be null there), while still being included if someone initialize their instance variables outside of that.

Not sure if we want to base our design on non-guaranteed optimizations like this, though? On the other hand, initialization code inside `declare_class!` should be _very_ linear with regards to `Allocated`, and OOM _is_ a very, very rare situation (in contrast to e.g. initialization failure), so perhaps it's fine to hide it a bit more from our API?

I guess I'll have to experiment with it.

madsmtm:
Another benefit is that also allows making convenience initializers every so slightly easier:
```rust
declare_class!(
   ...

    #[initConvenience]
    fn init_convenience(this: Allocated<Self>) -> Id<Self> {
        // Before
        let this = unsafe { msg_send![Some(this), init] };
        // After
        let this = unsafe { msg_send![this, init] };

        // Do something with `this`
        this.ivar = 42;
        this
    }
);
```

Which is important for `extern_methods!` and `declare_class!` being more symmetric (which in turn will make https://github.com/madsmtm/objc2/issues/455 easier to progress on).

madsmtm:
Sadly, messaging `init` methods with the wrong return `Id<R>` compared to `Allocated<T>` (i.e. `T != R`) now gives worse diagnostics. I'm fairly sure this can be fixed, but it requires refactoring how message sending works (i.e. instead of using a trait, move `MsgSendId` to be inherent methods on `RetainSemantics`. Actually this has potential to improve diagnostics a lot).

But I'm going to go through with it anyway, the diagnostic is not _that_ much worse.

PR #525, opened 2023-10-18
Add unstable-no-link feature

xiaopengli89:
Some frameworks may only be available on higher OS versions and some applications want to check at runtime, for example:

```rust
// Load MetalFX.framework
if AnyClass::get("MTLFXSpatialScalerDescriptor").is_none() {
    let path = ns_string!("/System/Library/Frameworks/MetalFX.framework");
    let Some(metal_fx_bundle) = NSBundle::bundleWithPath(path) else {
        println!("MetalFX.framework not found");
        return;
    };
    if !metal_fx_bundle.load() {
        println!("Load MetalFX.framework failed");
        return;
    }
}
```


crates/icrate/Cargo.toml
```diff
@@ -85,7 +85,10 @@ required-features = [
 ]
 
 [features]
-default = ["std", "apple"]
+default = ["std", "apple", "link"]
+
+# Link against framework, otherwise the correct framework should be load at runtime
+link = []
```
madsmtm, line 91: I think I'd prefer this feature to be called `no-link`, since it is a thing that basically only binaries would be concerned with enabling. E.g. a library like `winit` would not want to touch this feature at all, but users of `winit` might still want the functionality.

We could consider having features `no-link-Foundation`, `no-link-AppKit` and so on instead, since that way it would be less of a breaking change if some library started depending on e.g. `AppKit`.
Alternatively, it should be clearly documented that enabling the `no-link` feature opts out of certain SemVer guarantees.

crates/icrate/Cargo.toml
```diff
@@ -85,7 +85,10 @@ required-features = [
 ]
 
 [features]
-default = ["std", "apple"]
+default = ["std", "apple", "link"]
+
+# Link against framework, otherwise the correct framework should be load at runtime
```
madsmtm, line 90: Nit: Can you document the use-case you mentioned, as well as the related use-case in https://github.com/servo/core-foundation-rs/pull/608?

madsmtm (changes requested):
Thanks, I've been wanting this feature

crates/icrate/Cargo.toml
xiaopengli89: Done.

xiaopengli89:
The ci failed with negative_feature_names, but I don't know how to exempt it.

crates/icrate/Cargo.toml
```diff
@@ -87,6 +87,47 @@ required-features = [
 [features]
 default = ["std", "apple"]
 
+# Disable to link specific frameworks, which may not available on lower OS versions, or users don't want to load the frameworks at startup (dyld).
```
madsmtm, line 90: Sorry for being so slow to respond on this PR :/

I think I've changed my mind again on how I want this: We should go back to the single feature, but it should be called `unstable-no-link`, to signify that it should generally only be enabled by binary crates since it may break your build (since library crates are allowed to start depending on a framework in a minor release, and then they suddenly wouldn't be linked any more).

xiaopengli89: Done.

madsmtm:
Ok, last thing, promise this time:

Can your use-case be solved by `-weak_framework MetalFX` instead? E.g. is it enough to add the following to your crate (using a `nightly` from this week):
```rust
#![feature(link_arg_attribute)]

#[link(name = "-weak_framework", kind = "link-arg", modifiers = "+verbatim")]
#[link(name = "MetalFX", kind = "link-arg", modifiers = "+verbatim")]
extern "C" {}
```

xiaopengli89:
> Ok, last thing, promise this time:
> 
> Can your use-case be solved by `-weak_framework MetalFX` instead? E.g. is it enough to add the following to your crate (using a `nightly` from this week):
> 
> ```rust
> #![feature(link_arg_attribute)]
> 
> #[link(name = "-weak_framework", kind = "link-arg", modifiers = "+verbatim")]
> #[link(name = "MetalFX", kind = "link-arg", modifiers = "+verbatim")]
> extern "C" {}
> ```

Yeah, it works. I'll close the pr.

Issue #511, opened 2023-09-12
Using AtomicPtr as ivar


ilmai:
It looks to me like the constraint for using AtomicPtrs in ivars is wrong, or alternatively I'm misunderstanding something.

The equivalent of `*mut c_void` is `AtomicPtr<c_void>` but `Encode` is only implemented for `AtomicPtr<T: RefEncode>`, and `c_void` isn't `RefCode`. I think the correct constraint should be `*mut T: RefEncode`?

madsmtm:
The implementation on `AtomicPtr` is correct, the issue is that we're being a bit hacky around `c_void`, see https://github.com/madsmtm/objc2/blob/812a615bcd8647020016452565dcb3ab3c4d56a7/crates/objc2/src/encode.rs#L570-L572

Honestly, I should probably remove that hack, it's not really worth the inconsistency, and people who do FFI know not to use `&c_void` and `&mut c_void`.

In the meantime, you can use a wrapper like this:
```rust
#[repr(transparent)]
struct MyWrapper(AtomicPtr<c_void>);

unsafe impl Encode for MyWrapper {
    const ENCODING: Encoding = Encoding::Atomic(&Encoding::Pointer(&Encoding::Void));
}
```

PR #535, opened 2023-12-03
Remove `c_void` encoding hack

madsmtm:
This was in theory useful back when references to `c_void` were of higher likelihood, but nowadays I doubt it'll ever be an issue.

Fixes https://github.com/madsmtm/objc2/issues/511.

### DIFF
--- a/crates/test-ui/ui/msg_send_underspecified_error.stderr
+++ b/crates/test-ui/ui/msg_send_underspecified_error.stderr
@@ -3,24 +3,3 @@ error[E0282]: type annotations needed
   |
   |     let _: Result<(), _> = unsafe { msg_send![obj, error: _] };
   |            ^^^^^^^^^^^^^ cannot infer type
-
-error[E0283]: type annotations needed
- --> ui/msg_send_underspecified_error.rs
-  |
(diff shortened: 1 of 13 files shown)


crates/test-ui/ui/msg_send_underspecified_error.stderr
```diff
@@ -1,89 +1,5 @@
...
-  = note: required for `*mut *mut _` to implement `Encode`
-note: required by a bound in `send_message_error`
- --> $WORKSPACE/crates/objc2/src/__macro_helpers/msg_send.rs
-  |
-  |     unsafe fn send_message_error<A, E>(self, sel: Sel, args: A) -> Result<(), Id<E>>
-  |               ------------------ required by a bound in this associated function
-  |     where
-  |         *mut *mut E: Encode,
-  |                      ^^^^^^ required by this bound in `MsgSend::send_message_error`
-  = note: this error originates in the macro `$crate::__msg_send_helper` which comes from the expansion of the macro `msg_send` (in Nightly builds, run with -Z macro-backtrace for more info)
```
madsmtm, line 26: It may seem unfortunate that we're now missing this information, but in reality it's not actually helpful for the user.

Issue #266, opened 2022-09-08
How do we handle availability?


madsmtm:
In Apple's Objective-C headers, most classes and methods are annotated with an availability attribute such as `API_AVAILABLE(macos(10.13), ios(11.0), watchos(4.0), tvos(11.0))`; this is absolutely a _great_ idea (!!!), it very cleanly allows you to mark which APIs you may use, and which ones are not usable on your current deployment target.

We should have some way of doing the same as `@available` in Objective-C; however, since we are not a compiler like `clang` is, this is quite tricky!

A quick demonstration of what I want:
1. Declare an API which is only available on a specific deployment target
    ```rust
    #[cfg_available(macOS(10.10), ...)]
    #[sel(doSomething:)]
    fn doSomething(&self, arg: i32);
    ```
2. Prevent the user from using said API if their deployment target is not high enough:
    ```rust
    // MACOSX_DEPLOYMENT_TARGET=10.7
    obj.doSomething(32); // Fails (ideally at compile time, otherwise with debug assertions at runtime)
  
    // MACOSX_DEPLOYMENT_TARGET=10.10
    obj.doSomething(32); // Works
    ```
3. Allow the user (usually libraries) to use the API if they've verified that the target version is high enough (this uses a dynamic runtime check except if the deployment target is high enough).
    ```rust
    if available!(macOS(10.12, ...)) { // Anything higher than what `doSomething` needs
        obj.doSomething(32); // Works no matter the deployment target
    }
    ```

For this, [Contexts and capabilities](https://tmandry.gitlab.io/blog/posts/2021-12-21-context-capabilities/) come to mind, similar to how it would be [useful for autorelease pools](https://github.com/madsmtm/objc2/issues/86#issuecomment-1003371984), but alas, we can't do that yet, so this will probably end up as a debug assertions runtime check.

See also upstream https://github.com/SSheldon/rust-objc/issues/111, a WIP implementation can be found in https://github.com/madsmtm/objc2/pull/212.

madsmtm:
There are effectively two versions that affect availability and what we should do: the **deployment target** and the **SDK version**.

To illustrate, let's assume an API `fn foo() { ... }` that's introduced in some version `i`, later deprecated in some version `d`, and finally removed in some version `r`*.

| Deployment target | SDK version | `fn` declaration | `foo()` | `if_available(introduced_version_or_above) { foo() }` |
| --- | --- | --- | --- | --- |
| `..i` | `..i` | None | Fails | Fails |
[...]

PR #546, opened 2023-12-26
Add `ServiceManagement` framework

dongcarl:
Ref: https://developer.apple.com/documentation/servicemanagement?language=objc

Generated PR: https://github.com/madsmtm/icrate-generated/pull/9

-----

A note: When running `header-translator`, I was getting a panic at:
https://github.com/madsmtm/objc2/blob/5391a0df65ccb48f21fe64a79e18779f72c0af6e/crates/header-translator/src/main.rs#L280

I logged the filename and found that `header-translator` is trying to get the `ServiceManagement.h` file, which didn't exist in `library.files`.

The reason why it's trying to get `ServiceManagement.h` is because unlike most "umbrella" files in frameworks, the `ServiceManagement.h` file has a few `const CFStringRef` declarations in them.

This is perhaps also reflected in `ServiceManagement`'s modulemap, where the umbrella declaration is:
```
        umbrella "Headers"
```

Instead of something found in all other `icrate` supported Frameworks:

```
        umbrella header "CoreWLAN.h"
```

Since we need to skip all the declarations in `ServiceManagement.h` for `CoreFoundation` reasons, a workaround is just to delay unwrapping the file unless we parsed usable statements, which is implemented in this PR

-----

A question: The "Service Management" framework in general is supported on earlier macOS versions than the `SMAppService`, so which versions should I put in `crates/header-translator/translation-config.toml`? Also, which `unstable-frameworks-macos-*` should I put it under?


madsmtm:
CI is failing because you're using a newer version of Xcode than `icrate` is currently generated with.

> which versions should I put

Put the version of the framework; the supported versions of the framework will be automatically imported after we resolve #266.

> modulemap

Yeah, we don't support that at all, the plan is to integrate better with `clang` first.

dongcarl:
Seems like things are passing now

dongcarl:
Anything else needed here?

madsmtm (approved):
Nope, looks good, thanks!

Issue #500, opened 2023-09-03
Features that require `libc`


madsmtm:
We kinda need to figure out a general policy on how we handle features that require `libc` (or at least just linking to it).

There are at least two things I'd like to do:
- [x] Move `MainThreadMarker` into `objc2` (with a more efficient implementation, namely [`libc::pthread_main_np`](https://docs.rs/libc/0.2.146/x86_64-apple-darwin/libc/fn.pthread_main_np.html))
  - Although really convenient, unsure if this is desired to have in `objc2`? I feel like it belongs in some lower-level library, though we don't have that, so I guess it'll have to be here.
- [x] The things gated behind `malloc` support requires `libc::free`, I'd like to remove that feature gate to be able to verify more things in `debug_assertions`. Done in https://github.com/madsmtm/objc2/pull/547.

silvanshade:
Is there any real downside to using it?

I'm using it in the cxx-auto stuff FWIW, although that would only affect the header-translator.

madsmtm:
Main downside is compilation speed, it's an extra dependency (with a `build.rs` script).

I'd be fine with using it in `icrate`, though also there probably behind a feature flag, again because it's an added dependency.

madsmtm:
https://github.com/madsmtm/objc2/pull/616 changed the implementation of `MainThreadMarker` to use `pthread_main_np` and https://github.com/madsmtm/objc2/pull/625 moves it to `objc2`.

PR #547, opened 2023-12-27
Work towards removing `malloc_buf`

madsmtm:
`malloc_buf` is [unmaintained and unsound](https://github.com/rustsec/advisory-db/issues/1635), and may be marked as such in the Rustsec Advisory Database in the future, which would be bad for our users; so we should really migrate to something else.

The alternatives [`mbox`](https://github.com/kennytm/mbox), [`malloced`](https://github.com/nvzqz/malloced) and [`malloc-array`](https://github.com/notflan/rs-malloc-array) are either also unsound (usually failing to handle `0`-length arrays properly), or not popular/maintained enough for me to be comfortable with them as a dependency.

So let's just roll our own implementation, it's not really _that_ complex anyhow (since we don't handle allocation, only deallocation), and we should be able to swap it out with `Box<_, MallocAllocator>` in the future.

Part of https://github.com/madsmtm/objc2/issues/500.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ env:
   #
   # Note: The `exception` feature is not enabled here, since it requires
   # compiling C code, even if just running a `check`/`clippy` build.
-  INTERESTING_FEATURES: malloc,block,verify,unstable-private
+  INTERESTING_FEATURES: malloc,block,unstable-private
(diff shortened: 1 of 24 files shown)


madsmtm:
Note that this added `objc_sys::free`, to avoid the dependency on `libc`. I did first [investigate](https://users.rust-lang.org/t/can-i-rely-on-system-dealloc-calling-free/104467) if `System.dealloc` could be used for this, since that calls `libc::free` internally, but you probably can't rely on that.

PR #560, opened 2024-01-16
Update nighty Rust version in CI

madsmtm:


crates/objc2/src/rc/id_forwarding_impls.rs
```diff
@@ -21,6 +21,7 @@ use std::io;
 use super::Id;
 use crate::mutability::IsMutable;
 
+#[allow(clippy::unconditional_recursion)]
```
madsmtm, line 24: Opened https://github.com/rust-lang/rust-clippy/issues/12154.

Issue #566, opened 2024-01-20
Swift array count: expected return to have type code 'q', but found 'Q'


1tgr:
I'm exposing this Rust function to Swift:

```rust
#[no_mangle]
pub extern "C" fn time_series_eval(
    exprs: *mut NSArray<NSString>,
    error: &mut *mut NSError,
) -> *mut NSDictionary<NSString, NSArray> {
    let exprs = exprs.as_ref().unwrap().iter();
    // do something with exprs
}
```

I call it like this from the Swift side:
```swift
var error: NSError?
let result = time_series_eval(["abc"], &error)
```

But the Rust panics:
```
~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/icrate-0.1.0/src/generated/Foundation/NSEnumerator.rs:6:1:
invalid message send to -[Swift.__SwiftDeferredNSArray countByEnumeratingWithState:objects:count:]: expected return to have type code 'q', but found 'Q'
```

I think this is telling me that Swift arrays and objc2's arrays don't agree on signed vs unsigned?

madsmtm:
> I think this is telling me that Swift arrays and objc2's arrays don't agree on signed vs unsigned?

Rather, Swift is lazy and uses `Int` instead of the more correct type `UInt` here; the difference doesn't really matter to them, as arrays likely won't be able to ever get that large.

I've opened https://github.com/madsmtm/objc2/pull/567 to fix this, although I'm a bit unsure what the right approach is. In the meantime the easiest fix for you is probably to do one of:
- Patch in https://github.com/madsmtm/objc2/pull/567 locally (using the `patch.crates-io` Cargo.toml key).
- Re-implement the iteration yourself using `objectAtIndex:`.
- Disable debug assertions.

madsmtm:
I went with a Cargo feature flag called `relax-sign-encoding`, you should be able to use that instead once I release `objc2 v0.5.2`

madsmtm:
Would it be possible for you to give me a code example that miscompiles if I use `-> u8` in the signature instead of `-> i8` (or vice-versa)?

And similarly for `-> usize` and `-> isize`?

madsmtm:
Hmm, I can reproduce your result, but from my testing, it doesn't seem to produce differing results if I do `int8_t foo();`? So maybe what we're seeing here is more a problem with C's implicit integer casting rules?

In any case, the thing I'm really interested in is whether you know a case where zero/sign extension happens when taking/returning register-sized types?

[...]

PR #567, opened 2024-01-21
Ignore signed-ness in encoding

madsmtm:
Fixes https://github.com/madsmtm/objc2/issues/566.

Unsure if this is the correct path forwards.

Do signed-ness ever matter ABI-wise? The Rust docs [says](https://doc.rust-lang.org/nightly/std/primitive.fn.html) on ABI of `i32` vs `u32`:

> on some targets, the calling conventions for these types differ in terms of what they guarantee for the remaining bits in the register that are not used by the value.

But is that relevant for any platforms we target? And does it matter for types like `usize`/`isize`, which won't _have_ remaining bits left in any registers?

### DIFF
--- a/crates/objc2/CHANGELOG.md
+++ b/crates/objc2/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 * Added `Id::autorelease_ptr`.
+* Added the feature flag `"relax-sign-encoding"`, which when enabled, allows
+  using e.g. `NSInteger` in places where you would otherwise have to use
+  `NSUInteger`.
 
 ### Deprecated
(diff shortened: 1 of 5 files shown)


madsmtm:
Swift [doesn't make a distinction between `NSInteger` and `NSUInteger`](https://github.com/apple/swift/blob/24814c5e9da1552e3886bf380c53da1f10d8335a/lib/ClangImporter/MappedTypes.def#L154-L156) when importing methods, though [they do so when importing other signed/unsigned C types](https://github.com/apple/swift/blob/24814c5e9da1552e3886bf380c53da1f10d8335a/lib/ClangImporter/ClangAdapter.cpp#L338-L356), so maybe we should handle pointer-sized types specially here?

madsmtm:
I've decided to go with a feature flag `relax-sign-encoding` for now. In the future, we should probably restrict the encoding checks for `extern_methods!` in CI, and not on the user's machine

Issue #531, opened 2023-11-27
Generate KVO add/remove observer methods for NSProgress


knopp:
These methods are currently only generated for `NSSet` and `NSOrderedSet`. Would be nice to have it for other classes as well.

madsmtm:
KVO is generally ill-supported in `icrate` atm., definitely agree that more work is needed here!

Issue #501, opened 2023-09-04
Linking against private frameworks?


NyaomiDEV:
I wonder if it is inside the scope of this project to be able to link against private frameworks (specifically MediaRemote) and to use them in a mostly-type-safe way.

silvanshade:
There has been a little discussion around being able to use the [`header-translator`](https://github.com/madsmtm/objc2/tree/master/crates/header-translator) (the tool `icrate` uses to generate the Rust definitions from the Apple frameworks) for other frameworks. For example, see item 2 at https://github.com/madsmtm/objc2/issues/429.

So far there hasn't been a lot of work toward making this practical though. In principle, I believe it should be possible now, but it would require some modification to the `icrate` configuration and potentially to some parts of `header-translator`.

The situation should improve in the next few months hopefully, once there has been more progress made toward the new dependency scanning backend.

madsmtm:
At some point after https://github.com/madsmtm/objc2/issues/345 we'll probably figure out a good way for translating headers not provided via. Xcode, then you should be able to just run `header-translator` (or whatever it'll end up being named) on your header and get a Rust library that targets that.

madsmtm:
Oops, @silvanshade beat me to it, but agree with everything they said

Issue #309, opened 2022-12-23
Autogenerate documentation in `icrate`


madsmtm:
Opening this to track implementation work after https://github.com/madsmtm/objc2/pull/264.

Possibilities for having documentation in `icrate` include:
1. Links to https://developer.apple.com/
2. Grab documentation from the above and inline it
3. Convert code comments in the headers to documentation comments
4. Use [Xamarin's](https://github.com/xamarin/apple-api-docs)
5. ?

Or some combination of these.

For option 2, I did some preliminary work in https://github.com/madsmtm/objc2/issues/85#issuecomment-1236362033 (the files are available locally as well).

madsmtm:
Adding documentation for `weak` properties was done in 27ab005dbaefaafd0cacb4dc2d9682b784bd5e92, and linking to Apple's documentation was done in a8e99f6dc33fd9edcd6014aa782eb3370bd1728a.

madsmtm:
The auto-generated documentation is still pretty bad though, we need to either:
- Build a crate capable of properly parsing the semi-Sphinx syntax that Apple's headers use.
- Work to improve this parser in upstream `libclang`.

With the latter being my preference, but probably also being a lot harder.

madsmtm:
I've also added links to Apple's online documentation a lot of types, but this isn't yet done for function / methods, because it was unclear exactly how the links to those names were constructed.

But I see that Apple has recently changed what their links look like?

Sample functions to think about:
- `NSApplicationMain` -> `https://developer.apple.com/documentation/appkit/nsapplicationmain`
- `CFErrorCreate` -> `https://developer.apple.com/documentation/corefoundation/cferrorcreate(_:_:_:_:)`
- `CGContextConvertPointToUserSpace` -> `https://developer.apple.com/documentation/coregraphics/cgcontext/converttouserspace(_:)-3mtg3`
- `CGContextConvertRectToUserSpace` -> `https://developer.apple.com/documentation/coregraphics/cgcontext/converttouserspace(_:)-1hk5r`

And methods:
- `-[UIView initWithFrame:]` -> `https://developer.apple.com/documentation/uikit/uiview/init(frame:)`
- `-[UIView backgroundColor]` -> `https://developer.apple.com/documentation/uikit/uiview/backgroundcolor`

[...]

PR #574, opened 2024-02-04
Generate external categories

madsmtm:
Categories are a way for Objective-C code to add extra methods to a class. There are basically two ways we could handle these in `header-translator`:
1. Emit `extern_methods!` directly on the class (done currently).

    This breaks when the class is defined in a different library/framework than the category, because of Rust's coherence rules.
2. Create a trait, and implement that for the class (done in `uikit-sys`, and now by `extern_category!`).
  
    One problem here is that the category name is not part of the API in Objective-C, so we actually have no stable name to refer to. Relatedly, it's annoying to have to import category traits with confusing names all the time.

Both approaches have issues, so I've tried to strike a nice balance; we now emit `extern_methods!` when the category is defined in the same library/framework as the class, and `extern_category!` otherwise. This doesn't completely resolve the issues around categories not having a stable name, but it does somewhat alleviate them.

This allows us to fix https://github.com/madsmtm/objc2/issues/531, we now emit the missing key-value coding categories (and other such categories on `NSObject`). This should also help with splitting `icrate` up into multiple crates in the future, instead of it being a big monolithic one. Finally, this is required for `header-translator` to work on user-defined categories in the future (see https://github.com/madsmtm/objc2/issues/501).

CC @simlay as you've gone with the second approach in `bindgen`'s Objective-C support, wanted to hear your thoughts about this combined solution? I note that yours is currently better, since it implements the category trait for subclasses as well.

### DIFF
--- a/crates/header-translator/src/availability.rs
+++ b/crates/header-translator/src/availability.rs
@@ -139,11 +139,10 @@ impl Availability {
             _swift,
         }
     }
-}
 
-impl fmt::Display for Availability {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match &self.deprecated {
+    pub fn is_deprecated(&self) -> bool {
+        !matches!(
+            self.deprecated,
             Versions {
                 ios: None,
                 ios_app_extension: None,
@@ -153,7 +152,15 @@ impl fmt::Display for Availability {
                 watchos: None,
                 tvos: None,
                 visionos: None,
-            } => {
+            }
+        )
+    }
+}
+
+impl fmt::Display for Availability {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.deprecated {
+            _ if !self.is_deprecated() => {
                 // Not deprecated
             }
             Versions { .. } => {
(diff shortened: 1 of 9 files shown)


simlay:
> Both approaches have issues, so I've tried to strike a nice balance; we now emit `extern_methods!` when the category is defined in the same library/framework as the class, and `extern_category!` otherwise. This doesn't completely resolve the issues around categories not having a stable name, but it does somewhat alleviate them.

Yeah, I think this is a reasonable trade off. I think the negative to this approach is that debugging/understanding is harder. There's two possible ways a function is added to a rust-side class - importing the trait that the category implements (a class has a category from a different library/framework) vs not.

The ergonomics of including a bunch of traits just for a few specific functions was pretty rough.

Semirelated, I wish there was a good way to do source maps/links to apple docs for things.

madsmtm:
> > Both approaches have issues, so I've tried to strike a nice balance; we now emit `extern_methods!` when the category is defined in the same library/framework as the class, and `extern_category!` otherwise. This doesn't completely resolve the issues around categories not having a stable name, but it does somewhat alleviate them.
> 
> Yeah, I think this is a reasonable trade off. I think the negative to this approach is that debugging/understanding is harder. There's two possible ways a function is added to a rust-side class - importing the trait that the category implements (a class has a category from a different library/framework) vs not.

Thanks for the feedback; I'll try to document it at some point, tracking in https://github.com/madsmtm/objc2/issues/576.

> Semirelated, I wish there was a good way to do source maps/links to apple docs for things.

Well, there is no _good_ way, but there probably is a way, tracking in https://github.com/madsmtm/objc2/issues/309.

Issue #537, opened 2023-12-03
Split `icrate` into multiple crates


madsmtm:
We're currently hitting `crates.io`'s [limit of 300 features](https://blog.rust-lang.org/2023/10/26/broken-badges-and-23k-keywords.html), we should do something about that.

Maybe split the crate into multiple, e.g. `icrate-appkit`, `icrate-foundation`? Would also decrease the download size. Difficult because of coherence though.

madsmtm:
I've [requested an increase](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/5.20billion.20features/near/405687261) in the interim, though we should still solve this in a better way!

madsmtm:
`icrate v0.1.0` is now released, keeping this issue open to track reducing number of features

madsmtm:
We could of course just go back to having one `cfg` per framework, and then live with the worse compile-times until we figure something better out.

madsmtm:
Regarding splitting `icrate` into multiple smaller crates:

I think the coherence-problem is solvable, by properly introducing categories (in Swift, these are called "extensions"). In most cases, we will just emit categories as implementations directly on the type, as we're doing now. But if we encounter a category that is implemented on a type outside of the current library/framework, then we will emit a trait, and implement the trait for the type. This is a bit ugly, but we have to support categories if we want to support external crates anyhow.

Another problem is code-organization, as I'd still like to keep the generated files outside the main repository. Do we just use symlinks? I recall these being troublesome on Windows, will have to check how bad it is, but honestly, not a big issue if it's a bit more difficult for Windows users to clone the repository. The crates themselves won't have any symlinks, pretty sure Cargo prevents that, but will also have to verify that.

madsmtm:
Yeah, symlinks are a bit of trouble on Windows, but it's [solvable](https://stackoverflow.com/a/59761201/5203369). Cargo does copy files when running `cargo publish`/`cargo package`, but on Windows that of course [also depends on if the repo is correctly checked out](https://github.com/rust-lang/cargo/issues/5664).

madsmtm:
Other problems:

1. Frameworks sometimes reference each other in cyclic ways. E.g. `AppKit` references `Foundation`, while `Foundation` references `NSImage` from `AppKit`. I think these are rare enough that they can be resolved by hand.

[...]

PR #577, opened 2024-02-05
Use Cargo's new `[workspace.package]` fields

madsmtm:
See [the docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table).

This bump MSRV to 1.64, which should be fine, Winit is at 1.70 and `libc` is at 1.71.

This is done as a prerequisite for [splitting `icrate` into multiple crates](https://github.com/madsmtm/objc2/issues/537).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,8 +174,8 @@ jobs:
 
     - name: Use MSRV toolchain
       run: |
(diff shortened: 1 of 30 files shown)


madsmtm:
Superseded by deb95fcd34d9133f780d5f33ab4bae0604321406.

Issue #563, opened 2024-01-17
Consider removing (non-interior) mutability?


madsmtm:
So... I've kinda considered this problem many times, https://github.com/madsmtm/objc2/issues/265 contains some of the previous discussion and my head contains even more, but I feel like we're still not at an optimum.

Currently, `objc2` contains a lot of complex machinery to make classes able to opt-in to being `&mut`, and `icrate` uses this on `NSMutableString`, `NSMutableAttributedString`, `NSMutableArray<T>`, `NSMutableSet<T>`, `NSMutableDictionary<K, V>` and `NSEnumerator<T>` to make accessing the inner state of those types safe.

A few examples of current issues in `icrate` with this approach:
- `NSMutableAttributedString::mutableString` must return a reference bound to the attributed string, as otherwise you'd be able to create two `Id<NSMutableString>` to the same string. But this doesn't work with `Id<T>`, so we'd need a lifetime annotation in there somehow (or maybe resort to autorelease pools).
- `NSTreeNode::mutableChildNodes` has to remain `unsafe`, as we have no way of preventing calling that multiple times. Or maybe `NSTreeNode` also has to be mutable? But that in turn would push `NSTreeController` to being mutable, which in turn needs us to change `NSEditor`.
- While `NSThread::threadDictionary` has to remain `unsafe` because of thread-safety concerns, it cannot even be used from the owning thread safely.
- In user code, storing one of these mutable objects in a custom object needs to be done using a `RefCell`.

So while `&mut` does work for the select few types we've introduced it on, my main problem is that _this is not enough_! `&mut` is infectious, and we need to introduce it consistently, everywhere!

And that, in turn, is just not feasible (time-wise, Apple has a _lot_ of API surface)! Especially so with [our goal of making `header-translator` a user-accessible tool](https://github.com/madsmtm/objc2/issues/429), which means user-controlled code, and in turn means users would also have to add extra configuration to handle this for their mutable types, which is a tough ask.

Honestly, maybe we should abandon this endeavour, and instead go down the beaten path Swift has taken; do not try to do anything about mutability, i.e. mark `NSString`, `NSMutableString` and so on as `InteriorMutable`, and have all methods accept `&self`?

---

All of this is not to say that `&mut` doesn't have advantages; it does, very much so! And removing it does have several performance-related disadvantages, namely:
[...]

PR #587, opened 2024-03-05
Remove non-interior mutability from generic types

madsmtm:
Part of https://github.com/madsmtm/objc2/issues/563.

Remove non-interior mutability from NSEnumerator, NSArray, NSDictionary, NSSet and the mutable variants.

I've run the new `mut_while_iter` fuzz target for using AFL++ ~8 hours with 8 parallel processes, and using libFuzzer for ~7 hours with 6 parallel processes, to test whether Foundation's mutation detection is precise enough to catch all errors, such that we can declare it sound.

A concrete place where this will be useful is Winit, which stores a `RefCell<Retained<NSMutableAttributedString>>` inside of a subclass of `NSView` to track marked text, but the `RefCell` here should be completely unnecessary.


framework-crates/objc2-foundation/src/util.rs
```diff
@@ -11,88 +10,21 @@ pub(crate) fn retained_ptr_cast<T: ?Sized>(objects: *mut Retained<T>) -> *mut No
...
-pub(crate) unsafe fn mutable_collection_retain_removed<T>(obj: &T) -> Retained<T>
-where
-    T: Message,
-{
-    // SAFETY: The collection had mutable ownership over the object, and since
-    // the object will never again be accessed from the collection, we can
-    // convert it to `Retained<T>`.
+pub(crate) fn retain<T: Message>(obj: &T) -> Retained<T> {
+    // SAFETY: TODO
     unsafe { Retained::retain(obj as *const T as *mut T).unwrap_unchecked() }
```
madsmtm, line 29: Currently unsound, but won't be after we're done removing mutability.

PR #588, opened 2024-03-14
Update nightly Rust version

madsmtm:
Assembly test changes are:
- Smaller codegen for `Option::unwrap`
- Different register usage, especially around `CachedSel`. Perhaps slightly worse, but unsure, I suspect it might be a correctness fix.
- Some changes to `Box` creation and destruction.

UI test changes are:
- Added "which is required by", generally an improvement.
- "the following other types implement trait" suggestions are shown in fewer situations.


madsmtm:
Blocked by https://github.com/KyleMayes/clang-rs/issues/47.

Issue #558, opened 2024-01-14
Rename icrate features to end with `_core`


madsmtm:
Rename `Foundation` to `Foundation_core` (and similarly for the other features).

mwcampbell:
Would this be a precursor to being able to build icrate with fewer bindings for large frameworks like AppKit? The AppKit feature currently enables several bindings that AccessKit, for example, doesn't need.

madsmtm:
No, the proper way to do that is to do `cfg`s based on file names instead.

I remember trying to make that work before, but it's quite difficult to do with Libclang, I'm hoping to get to it once we get further with https://github.com/madsmtm/objc2/issues/345.

Issue #482, opened 2023-08-10
"StoreKit" feature depends on "AppKit", which is not available on iOS


wooden-worm:
I'm trying to use `icrate` on iOS. My goals are simple: showing IAPs and allowing users to purchase them. However, `icrate` with `StoreKit` feature fails to build for iOS due to its dependency on `AppKit` https://github.com/madsmtm/objc2/blob/3d9b6864af05592fc18b1ab27b3da9ed3896bc53/crates/icrate/Cargo.toml#L4582C1-L4585C2

Proposal: remove the dependency on `AppKit`. Users who need it can easily enable `AppKit`.

silvanshade:
Yeah, this is not ideal.

The goal is eventually to have the header translator that generates the Rust code from the Objective-C framework headers do a platform-specific analysis (e.g., for macOS vs iOS vs watchOS, ...) that would catch this sort of thing and only expose the correct features and sub-modules per-platform.

But right now that functionality is not really implemented and the actual backend logic for the translator is in the process of moving to a new, more robust approach, so it may be awhile before that functionality lands (probably at least another month or two, and that may be optimistic).

You can read a little about it here:

- https://github.com/madsmtm/objc2/issues/408
- https://github.com/madsmtm/objc2/issues/345#issuecomment-1599406283 (re: "diffing engine")

A one-off fix to address this specific case might be possible but the difficulty is in maintaining that change throughout the header generation process. There isn't really tool support for that now so it would have to be re-implemented by hand each time the generator is re-run.

Probably the most practical solution at the moment would be to maintain a local fork of your own with those changes that you need and then you could rebase your fork on top of whatever updates are released through `icrate` as needed.

Hopefully it will be possible to handle this more directly once the new binding generator starts to become usable.

bysensa:
Any updates here?

madsmtm:
https://github.com/madsmtm/objc2/pull/592, which will hopefully land soon, will fix this by making the AppKit dependency optional

bysensa:
> #592, which will hopefully land soon, will fix this by making the AppKit dependency optional

Awesome, thank you

Issue #408, opened 2023-01-29
iOS/tvOS/watchOS support in `icrate`


madsmtm:
My plan was to do something like:
1. Run the "collection" part of `header-translator` (e.g. `main.rs::parse_sdk`) for each platform
2. Compare the output against each other:
  - For each framework, if it exists on one platform but not the others, add it to the list of frameworks, but with appropriate `availability` attribute that will end up adding a `cfg` attribute on it
  - For each file, --||--
  - On each platform, put all statements into some sort of `HashMap` with "top-level" information (name of stmt, which generics a class has, and so on (but not which methods a class has implemented))
    - Use this to figure out which statements are on which some platforms
      - Add that to the final output, with relevant `cfg` attributes (ideally in the same order as the input file)
  - Similarly for each method in classes/protocols/categories
3. Output the combined result

The result is that we would only have to distribute one crate, and that crate would have the correct `cfg` attributes, which is _very_ useful for the documentation on docs.rs (since it could directly show us which things are available on which platforms).

As an example, let's take `NSPointerArray`: On this, there are five methods on macOS, and three on iOS, `pointerArrayWithStrongObjects` and `pointerArrayWithWeakObjects` are only available on iOS. So the result we'd like is that these two got a `#[cfg(target_os = "macos")]` on them.

Berrysoft:
May I ask the current progress?

madsmtm:
I haven't done much to further progress here. I think the basic idea is still sound, but I got kinda stuck on wanting to figure out a nice non-O(n^2) way to merge multiple files, but I guess it probably won't be that much of a problem in practice if we just define some sort of merge ordering.

Berrysoft:
I don't think it's an O(n^2). The merge order should not influence the merge result. Therefore it's O(n).

PR #592, opened 2024-03-19
Split `icrate` into one crate per framework

madsmtm:
Fixes https://github.com/madsmtm/objc2/issues/537.
Fixes https://github.com/madsmtm/objc2/issues/558.
Fixes https://github.com/madsmtm/objc2/issues/482.
Required for https://github.com/madsmtm/objc2/issues/408.

This should also improve compile times a bit, as Cargo can parallelize it much more, and it doesn't suffer from long download times.

The submodule path has been changed, in case git doesn't really show it you can view https://github.com/madsmtm/objc2-generated/commit/cf3442835f6121edbad2336eae871321cabc83a4 for the changed commit.

I'm fairly sure that the way we've chosen to do features won't turn into a breaking change, at the very least, `#import <WebKit/WebView.h>` makes the same, if not more types available to Objective-C, so it really shouldn't.

TODO:
- [x] ~Rename features to not contain the framework name~ Will do this in a follow-up PR.
- [x] Remove all mentions of `icrate`
- [x] Run full CI
- [x] ~Set up something to release these crates~ Will do this in a follow-up PR.


MediaEnhanced:
I tried this branch out with a very basic NSApplication, one NSWindow, slightly custom NSEvent, example and could only get it to compile (and run) after performing the following find and replaces over the generated files:
`use crate::AppKit::*; -> use crate::*;`
`use crate::CoreData::*; -> use objc2_core_data::*;`
`use crate::Foundation::*; -> use objc2_foundation::*;`
`use crate::common::*; -> use objc2_helpers::*;`

Not sure if that part in the "header-translator" has to be updated as this branch is WIP or if I setup something incorrectly. I also needed to remove the function feature restrictions for certain cross-framework restricted functions such as: `NSWindow::initWithContentRect_styleMask_backing_defer`
and others even though my Cargo.toml had the following:

    objc2 = { path = "src/icrate/crates/objc2" }
    objc2-foundation = { path = "src/icrate/framework-crates/objc2-foundation", features = [
	"Foundation_NSString",
	"Foundation_NSGeometry",
	"Foundation_NSNotification",
	"Foundation_NSObject",
	"Foundation_NSDate",
	"Foundation_NSThread",
	"Foundation_NSRunLoop",
	"Foundation_NSObjCRuntime",
	] }
    objc2-app-kit = { path = "src/icrate/framework-crates/objc2-app-kit", features = [
	"AppKit_NSApplication",
	"AppKit_NSRunningApplication",
	"AppKit_NSGraphics",
	"AppKit_NSGraphicsContext",
	"AppKit_NSEvent",
	"AppKit_NSWindow",
	"AppKit_NSResponder",
	] }
This branch did seem to build faster on my 2010 iMac compared to the master branch!

madsmtm:
The development here has been... quite haphazard. Will try to get it into a better state soon!

But I suspect that the issue you've been hitting is that git doesn't really like changing the path of a submodule, so you have to somehow tell it to re-initialize it - which I did manually by moving some directories, but there probably is a proper way of doing it.

PR #612, opened 2024-05-03
clippy: Enable `semicolon_if_nothing_returned`

waywardmonkeys:
One part of this required a change other than semicolons to fix type inferencing in `crates/objc2/src/__macro_helpers/declared_ivars.rs`.

### DIFF
--- a/crates/objc2/benches/autorelease.rs
+++ b/crates/objc2/benches/autorelease.rs
@@ -10,7 +10,7 @@ const BYTES: &[u8] = &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
 fn empty() {}
 
 fn pool_cleanup() {
-    autoreleasepool(|_| {})
+    autoreleasepool(|_| {});
 }
 
 fn class() -> &'static AnyClass {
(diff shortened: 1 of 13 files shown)


madsmtm (commented):
Thanks, although a lot of these are of the form:
```rust
unsafe {
    my_unsafe_expr; // <- Added semicolon
}; // Now redundant
```

The `unsafe` block is explicitly scoped for the unsafe expression inside of it, so I feel that this is a regression?

At least, if the expression were to change to become shorter, then `rustfmt` would be able to format it as:
```rust
unsafe { my_unsafe_expr };
```

waywardmonkeys:
Sorry, I was away for the weekend on a trip. I understand a reasoning (not sure if it is "the reasoning") for having the semicolons in those `unsafe` blocks (since they're like any other block...)

I could edit this PR to not enable the lint and not include the unsafe blocks. This would still be an improvement, I think, since there's a bunch of missing semicolons that don't involve unsafe blocks.

madsmtm:
> I could edit this PR to not enable the lint and not include the unsafe blocks. This would still be an improvement, I think, since there's a bunch of missing semicolons that don't involve unsafe blocks.

That'd be nice, please do so!

waywardmonkeys:
@madsmtm Updated. See what you think!

madsmtm (approved):
Yeah, this does make the code clearer, thanks!

PR #613, opened 2024-05-04
Add support for Virtualization framework

retrage:
This PR proposes to add Virtualization Framework support for macOS 12.0 or later. I'm developing a [Virtualization Framework-based VMM](https://github.com/retrage/akari) in Rust that deeply depends on [the fork of objc2](https://github.com/retrage/objc2/tree/icrate-virtualization-clean) with Virtualization Framework support.

According to [the official document](https://developer.apple.com/documentation/virtualization?language=objc), macOS 11.0 starts supporting the framework, and macOS 12.0 has additional APIs to support macOS guests. Since my VMM is designed to run macOS guests, the `objc2-virtualization` must support the APIs for macOS 12.0 or later. So I have hardcoded the minimum supported macOS version to 12.0 in `framework-crates/objc2-virtualization/translation-config.toml`. This may not be appropriate for other use cases.

This PR depends on https://github.com/madsmtm/objc2-generated/pull/10.


framework-crates/objc2-virtualization/translation-config.toml
```diff
@@ -0,0 +1,16 @@
+framework = "Virtualization"
+crate = "objc2-virtualization"
+required-dependencies = ["objc2-app-kit", "objc2-foundation"]
+macos = "12.0"
```
madsmtm, line 4: > Since my VMM is designed to run macOS guests, the `objc2-virtualization` must support the APIs for macOS 12.0 or later. So I have hardcoded the minimum supported macOS version to 12.0 in `framework-crates/objc2-virtualization/translation-config.toml`. This may not be appropriate for other use cases.

You should specify the version that it says on the front page, i.e. macOS 11.0. Newer APIs will still be available, the specified version in `translation-config.toml` is mostly for CI.

framework-crates/objc2-virtualization/translation-config.toml
```diff
@@ -0,0 +1,16 @@
+framework = "Virtualization"
+crate = "objc2-virtualization"
+required-dependencies = ["objc2-app-kit", "objc2-foundation"]
```
madsmtm, line 3: From what I can tell, a lot of the functionality does not require AppKit, so please leave that out the required list.

See [this doc](https://docs.rs/objc2/0.5.1/objc2/topics/about_generated/cargo_features/index.html) for more information on dependency features.

madsmtm (changes requested):
Thanks! Please make CI pass, I think you need to commit the update to the submodule to this branch as well, so that `git` can properly track it. Feel free to say if you need help doing so!

framework-crates/objc2-virtualization/Cargo.modified.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[features]
+default = ["std", "apple"]
+
+# Currently not possible to turn off, put here for forwards compatibility.
+std = ["alloc", "objc2/std", "block2?/std", "objc2-foundation/std"]
+alloc = ["objc2/alloc", "block2?/alloc", "objc2-foundation/alloc"]
+
+# Runtime selection. See `objc-sys` for details.
+apple = ["objc2/apple", "block2?/apple", "objc2-foundation/apple"]
```
madsmtm, line 17: All of this should only be necessary when allowing GNUStep

framework-crates/objc2-virtualization/translation-config.toml
retrage: Thanks. Removed from `requred-dependencies`.

framework-crates/objc2-virtualization/Cargo.modified.toml
retrage: Removed in [039aec0](https://github.com/madsmtm/objc2/pull/613/commits/039aec0d14717fc9aaca99a1e707ed453c5a8cf2).

retrage:
I noticed that some of the APIs are only available for aarch64 (they have `#ifdef __arm64__` guards). How can I deal with them? Is it OK to include them for all target architectures, or do they need to be guarded with `#[cfg(target_arch = "aarch64")]`?

retrage:
The "Verify header translator output" should succeed if the `header-translator` target is changed as follows:

```patch
diff --git a/crates/header-translator/src/main.rs b/crates/header-translator/src/main.rs
index e0003505..7be0c427 100644
--- a/crates/header-translator/src/main.rs
+++ b/crates/header-translator/src/main.rs
@@ -90,7 +90,7 @@ fn main() -> Result<(), BoxError> {
         // These are found using the `get_llvm_targets.fish` helper script
         let llvm_targets: &[_] = match &sdk.platform {
             Platform::MacOsX => &[
-                "x86_64-apple-macosx10.12.0",
+                "arm64-apple-macosx10.12.0",
                 // "arm64-apple-macosx11.0.0",
                 // "i686-apple-macosx10.12.0",
             ],
```

madsmtm:
Hmm, I'm a bit weary about doing so, but I think it'll be fine for now, please commit that change as well.

In the future I'm planning to automatically generate `cfg`-guards for target-specific APIs, but it's a hard problem.

retrage:
[25d88ee](https://github.com/madsmtm/objc2/pull/613/commits/25d88ee64280f81d1dadcb20179c774a9b3f4c0a) switches the target.

madsmtm (approved):
Thanks!

PR #619, opened 2024-05-21
New versions

madsmtm:
Release checklist:
- [x] The branch is named `new-versions`, such that the full CI will run.
    - `objc2-proc-macros`
    - `objc-sys`
    - `objc2-encode`
    - `objc2`
    - `block2`
    - Framework crates
- Local tests have been run (see `helper-scripts/test-local.fish`):
    - [x] macOS 10.14.6 32bit
    - [x] iOS 9.3.6, 1st generation iPad Mini

Post merge:
- [ ] A tag is created on the merge commit for each new version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,7 +57,7 @@ checksum = "cf4b9d6a944f767f8e5e0db018570623c85f3d925ac718db4e06d0187adb21c1"
 
 [[package]]
 name = "block2"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "objc2",
 ]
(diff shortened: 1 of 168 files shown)


madsmtm:
Wow, this one was a mess... I hit an issue with the version of Cargo I was using (several times) which prevented it from transforming the `lints.workspace = true` field before publishing, and then I also had to fork `cargo-release` to allow releasing more than 30 crates at a time, and `objc2-ui-kit` hadn't gotten its maximum features bumped, and so on... But, it's mostly done for now, I _really_ hope the next one's going to be easier.

Issue #77, opened 2021-11-22
Add Dispatch / `libdispatch`


madsmtm:
Reasoning:
- Better main thread support.
- `objc2` encoding support (allowing use in framework crates).
- [`async` support](https://github.com/madsmtm/objc2/issues/279)?
- Optimizations (such as the null-ptr optimization, and others).

Desired name: `dispatch2`, since it fits fairly well with the naming scheme (`objc2`, `block2` and `dispatch2` would all be successors to `objc`, `block` and `dispatch`, and would all be libraries internally, as compared to the rest of the crates we expose, which are all frameworks). I've asked for the name [here](https://github.com/marysaka/dispatch2/issues/2), should remember to not close this issue until that repo is archived too.

Unsure how auto-generated we can feasibly be here? But the library is comparatively small compared to most frameworks, so it's probably possible to generate an initial starting point, and then refine later on.

Upstream `dispatch` issues:
- [ ] https://github.com/SSheldon/rust-dispatch/issues/4
- [x] https://github.com/SSheldon/rust-dispatch/issues/7
- [ ] https://github.com/SSheldon/rust-dispatch/issues/14
- [x] https://github.com/SSheldon/rust-dispatch/issues/15
- [ ] https://github.com/SSheldon/rust-dispatch/issues/20

Upstream `dispatch` pull requests to integrate:
- [ ] https://github.com/SSheldon/rust-dispatch/pull/6
- [ ] https://github.com/SSheldon/rust-dispatch/pull/11
- [ ] https://github.com/SSheldon/rust-dispatch/pull/17
- [x] https://github.com/SSheldon/rust-dispatch/pull/21
- [ ] https://github.com/SSheldon/rust-dispatch/pull/22

Integrate ideas from:
- [ ] [`dispatch`](https://github.com/SSheldon/rust-dispatch)
- [ ] [`fruity`'s `dispatch`](https://github.com/nvzqz/fruity/tree/main/src/dispatch)
- [ ] [`dispatchr`](https://github.com/drewcrawford/dispatchr)
- [x] [`dispatch2`](https://github.com/marysaka/dispatch2)
- [ ] [cidre's `dispatch`](https://github.com/yury/cidre/tree/8227b566dfe08063a58dd3e4f557e6582cee7770/cidre/src/dispatch)

Other TODOs:
- [x] Use in framework crates.
- [ ] Refactor such that we have less ugly types lying around.

silvanshade:
[...]

PR #622, opened 2024-05-28
Add SystemExtensions framework / `objc2-system-extensions` crate

PaulDance:
Hi!

This proposes to add [`SystemExtensions`](https://developer.apple.com/documentation/systemextensions) as a new framework crate to `objc2`'s collection. It is a macOS 10.15+ -only framework. The `generated` directory is available [here](https://github.com/PaulDance/objc2-generated/tree/system-extensions/SystemExtensions).

It is relatively lightweight: only one module with a few items that felt standard Obj-C to me. The translation was thus not too hard, as only libdispatch required a few things to be skipped.

However, as the concerned APIs are at the center of framework, it is still a bit bothersome. #77 would therefore be nice for this. In the meantime, I guess the objects can still be somehow built with the generated bindings and the skipped methods used through `msg_send!` with the `dispatch` crate's types at least?

Cheers,
Paul.


PaulDance:
The [error in the Compiler UI step](https://github.com/madsmtm/objc2/actions/runs/9268413150/job/25497105241?pr=622) is a bit weird: it seems to only be movement while the message is basically the same. Is it something that has yet to be updated on `master` or should I update it here?

madsmtm:
I've attempted to fix the CI in 6f6ec78db700ead78a87edab30be9fe533da5821, please rebase, otherwise it looks good!

Issue #442, opened 2023-04-25
Create blocks with BLOCK_HAS_SIGNATURE


goffrie:
https://github.com/madsmtm/objc2/blob/dd6c804d1e189fd8c9b313d8c1901ee9e7286551/crates/block2/src/concrete_block.rs#L189

Some newer Apple APIs appear to fail in unexpected ways when called with blocks that don't have signatures. It would be useful to at least have some way of manually specifying a signature, though unsafe. Generating it automatically would be even better of course, though it's an interesting question how to actually do so.

madsmtm:
I have parts of the code for generating the encodings in a `const` context from a `Encoding::ENCODING`, but we're missing a few Rust features to make that a reality (see https://github.com/madsmtm/objc2/pull/70).

But we can definitely add a way to do it in an `unsafe` way for now.

Do you have an example of a few APIs that fail, so that I can add them to the test suite when implementing the feature?

goffrie:
The ones I know of are `NSFileProviderManager`'s `reimportItemsBelowItemWithIdentifier:completionHandler:` and `waitForStabilizationWithCompletionHandler:`, but I'm not sure how easy it would be to set up a test for those since it requires a FileProvider domain to be set up and everything.

madsmtm:
Thanks for the examples, I haven't worked with `FileProvider` myself, so probably the reason why I haven't hit this particular snag myself yet.

I probably won't have the time to implement a function for creating blocks with encodings at the moment (I'm out of the country, and my focus lies elsewhere), but I will happily review and accept a PR that does it (and I could mentor an implementer as well).

PaulDance:
I seem to have hit this limitation as well. The concerned framework is [`NetworkExtension`](https://developer.apple.com/documentation/networkextension?language=objc), and the specific API is [`NEFilterDataProvider`'s `applySettings:completionHandler`](https://developer.apple.com/documentation/networkextension/nefilterdataprovider/3181998-applysettings?language=objc).

The process crashes on the uncaught exception `Block was not compiled using a compiler that inserts type information about arguments`. The full error can be seen below.

<details><summary>Details</summary>
<p>

```
<NSXPCConnection: 0x600001528000> connection from pid 298 on anonymousListener or serviceListener: Warning: Exception caught during invocation of selector startFilterWithOptions:completionHandler:, dropping incoming message and invalidating the connection.
[...]

PR #636, opened 2024-07-05
Manual block encodings

PaulDance:
Hey @madsmtm!

This is an attempt to contribute to #442, by adding the ability to manually provide a block encoding to its constructor so that it sets `BLOCK_HAS_SIGNATURE` and saves the string properly.

Included:
 * New `StackBlock::with_encoding`.
 * New `RcBlock::with_encoding`.
 * `block2::traits::ManualBlockEncoding` is the smuggling trait.
 * String check on `cfg(debug_assertions)`: a bit too strict for now, but can be relaxed later on by implementing a dedicated parser and equivalence checking.
 * Updated and new tests.

Cheers,
Paul.

### DIFF
--- a/crates/block2/Cargo.toml
+++ b/crates/block2/Cargo.toml
@@ -47,6 +47,11 @@ unstable-private = []
 [dependencies]
 objc2 = { path = "../objc2", version = "0.5.2", default-features = false }
 
+[dev-dependencies.objc2-foundation]
+path = "../../framework-crates/objc2-foundation"
+default-features = false
+features = ["NSError"]
+
 [package.metadata.docs.rs]
 default-target = "aarch64-apple-darwin"
 features = ["unstable-private"]
--- a/crates/block2/src/stack.rs
+++ b/crates/block2/src/stack.rs
@@ -10,9 +10,11 @@ use core::ptr::{self, NonNull};
 use objc2::encode::{EncodeArguments, EncodeReturn, Encoding, RefEncode};
 
 use crate::abi::{
-    BlockDescriptor, BlockDescriptorCopyDispose, BlockDescriptorPtr, BlockFlags, BlockHeader,
+    BlockDescriptor, BlockDescriptorCopyDispose, BlockDescriptorCopyDisposeSignature,
+    BlockDescriptorPtr, BlockDescriptorSignature, BlockFlags, BlockHeader,
 };
 use crate::debug::debug_block_header;
+use crate::traits::{ManualBlockEncoding, ManualBlockEncodingExt, NoBlockEncoding, UserSpecified};
 use crate::{ffi, Block, IntoBlock};
 
 /// An Objective-C block constructed on the stack.
@@ -142,7 +144,12 @@ impl<'f, A, R, Closure: Clone> StackBlock<'f, A, R, Closure> {
     };
 }
 
-impl<'f, A, R, Closure> StackBlock<'f, A, R, Closure> {
+impl<'f, A, R, Closure> StackBlock<'f, A, R, Closure>
+where
+    A: EncodeArguments,
+    R: EncodeReturn,
+    Closure: IntoBlock<'f, A, R> + Clone,
+{
     /// Construct a `StackBlock` with the given closure.
     ///
     /// Note that this requires [`Clone`], as a C block is generally assumed
@@ -154,22 +161,100 @@ impl<'f, A, R, Closure> StackBlock<'f, A, R, Closure> {
     ///
     /// [`RcBlock::new`]: crate::RcBlock::new
     #[inline]
-    pub fn new(closure: Closure) -> Self
+    pub fn new(closure: Closure) -> Self {
+        // SAFETY: no encoding is given.
+        unsafe { Self::maybe_encoded::<NoBlockEncoding<A, R>>(closure) }
+    }
+
+    /// Constructs a new [`StackBlock`] with the given function and encoding
+    /// information.
+    ///
+    /// Some particular newer-ish Apple Objective-C APIs expect the block they
+    /// are given to be created with encoding information set in the block
+    /// object itself and crash the calling process if they fail to find it,
+    /// which renders them pretty much unusable with only [`Self::new`] that
+    /// currently does not set such encoding information. This is for example
+    /// the case in [`FileProvider`] for [`NSFileProviderManager`]'s
+    /// [`reimportItemsBelowItemWithIdentifier:completionHandler:`] and
+    /// [`waitForStabilizationWithCompletionHandler:`], but also in
+    /// [`NetworkExtension`] for [`NEFilterDataProvider`]'s
+    /// [`applySettings:completionHandler`]. A complete list of such APIs may
+    /// not be easily obtained, though.
+    ///
+    /// This encoding information string could technically be generated at
+    /// compile time using the generic parameters already available to
+    /// [`Self::new`]. However, doing so would require some constant evaluation
+    /// features that are yet to be implemented and stabilized in the Rust
+    /// compiler. This function is therefore exposed in the meantime so users
+    /// may still be able to call the concerned APIs by providing the raw
+    /// encoding information string themselves, thus obtaining a block
+    /// containing it and working with these APIs.
+    ///
+    /// The same requirements as [`Self::new`] apply here as well.
+    ///
+    /// [`FileProvider`]: https://developer.apple.com/documentation/fileprovider?language=objc
+    /// [`NSFileProviderManager`]: https://developer.apple.com/documentation/fileprovider/nsfileprovidermanager?language=objc
+    /// [`reimportItemsBelowItemWithIdentifier:completionHandler:`]: https://developer.apple.com/documentation/fileprovider/nsfileprovidermanager/reimportitems(below:completionhandler:)?language=objc
+    /// [`waitForStabilizationWithCompletionHandler:`]: https://developer.apple.com/documentation/fileprovider/nsfileprovidermanager/waitforstabilization(completionhandler:)?language=objc
+    /// [`NetworkExtension`]: https://developer.apple.com/documentation/networkextension?language=objc
+    /// [`NEFilterDataProvider`]: https://developer.apple.com/documentation/networkextension/nefilterdataprovider?language=objc
+    /// [`applySettings:completionHandler`]: https://developer.apple.com/documentation/networkextension/nefilterdataprovider/3181998-applysettings?language=objc
+    ///
+    /// # Safety
+    ///
+    /// The raw encoding string given through `E` must be correct with respect
+    /// to the given closure's argument and return types: see
+    /// [`ManualBlockEncoding`].
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use std::ffi::CStr;
+    /// # use block2::{Block, ManualBlockEncoding, StackBlock};
+    /// # use objc2_foundation::NSError;
+    /// #
+    /// struct MyBlockEncoding;
+    /// unsafe impl ManualBlockEncoding for MyBlockEncoding {
+    ///     type Arguments = (*mut NSError,);
+    ///     type Return = i32;
+    ///     const ENCODING_CSTR: &'static CStr = cr#"i16@?0@"NSError"8"#;
+    /// }
+    ///
+    /// let my_block = unsafe {
+    ///     StackBlock::with_encoding::<MyBlockEncoding>(|_err: *mut NSError| {
+    ///         42i32
+    ///     })
+    /// };
+    /// assert_eq!(my_block.call((std::ptr::null_mut(),)), 42);
+    /// ```
+    #[inline]
+    pub unsafe fn with_encoding<E>(closure: Closure) -> Self
     where
-        A: EncodeArguments,
-        R: EncodeReturn,
-        Closure: IntoBlock<'f, A, R> + Clone,
+        E: ManualBlockEncoding<Arguments = A, Return = R>,
+    {
+        // SAFETY: supposed to be upheld by the caller.
+        unsafe { Self::maybe_encoded::<UserSpecified<E>>(closure) }
+    }
+
+    #[inline]
+    unsafe fn maybe_encoded<E>(closure: Closure) -> Self
+    where
+        E: ManualBlockEncodingExt<Arguments = A, Return = R>,
     {
         let header = BlockHeader {
             #[allow(unused_unsafe)]
             isa: unsafe { ptr::addr_of!(ffi::_NSConcreteStackBlock) },
-            // TODO: Add signature.
-            flags: BlockFlags::BLOCK_HAS_COPY_DISPOSE,
+            flags: BlockFlags::BLOCK_HAS_COPY_DISPOSE
+                | if !E::IS_NONE {
+                    BlockFlags::BLOCK_HAS_SIGNATURE
+                } else {
+                    BlockFlags::EMPTY
+                },
             reserved: MaybeUninit::new(0),
             invoke: Some(Closure::__get_invoke_stack_block()),
             // TODO: Use `Self::DESCRIPTOR_BASIC` when `F: Copy`
             // (probably only possible with specialization).
-            descriptor: BlockDescriptorPtr {
+            descriptor: if E::IS_NONE {
                 // SAFETY: The descriptor must (probably) point to `static`
                 // memory, as Objective-C code may assume the block's
                 // descriptor to be alive past the lifetime of the block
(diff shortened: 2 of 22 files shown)


madsmtm:
> dangling pointers

Right, `&Self::DESCRIPTOR_WITH_CLONE` works because it is [promoted](https://doc.rust-lang.org/reference/destructors.html#constant-promotion) to a `static`, but that can't happen if we construct the descriptor on the stack. I've documented this somewhat in 893d0441b4132a4f01e7ca2e0007746ee7eb58bd.

I think my preferred solution (for now) would be something like the following instead (similar to your proposal in https://github.com/madsmtm/objc2/issues/442#issuecomment-2199966105, just with an extra trait):
```rust
unsafe trait BlockEncodingHelper { // Bikeshed name
    type Arguments: EncodingArguments;
    type Return: EncodingReturn;
    const ENCODING_CSTR: &'static CStr;
}

impl Block {
    pub fn with_encoding<'f, A, R, Closure, Helper>(
        _helper: Helper,
        closure: Closure,
    ) -> Self
    where
        Helper: BlockEncodingHelper<Arguments = A, Return = R>,
        // ...
    { ... }
}

// Usage
struct Helper;

impl BlockEncodingHelper for Helper {
    type Arguments = (i32, Bool);
    type Return = u8;
    const ENCODING_CSTR: &'static CStr = c"whatever_the_encoding_is_I_have_not_checked_it";
}

let block = RcBlock::with_encoding(Helper, |arg1, arg2| {
    if arg2.as_bool() {
        arg1 as u8
    } else {
        0
    }
});
```

PaulDance:
> Right, `&Self::DESCRIPTOR_WITH_CLONE` works because it is [promoted](https://doc.rust-lang.org/reference/destructors.html#constant-promotion) to a static

Ah yes, I was suspecting something of the sort was happening behind the scenes, but did not know the exact underlying mechanism. Thanks for the link.

> I think my preferred solution (for now) would be something like the following instead

That seems like it could work, yes thanks. I'll try it soon and see if the promotion works in this case too.

crates/block2/Cargo.toml
```diff
@@ -59,6 +59,12 @@ unstable-private = []
 [dependencies]
 objc2 = { path = "../objc2", version = "0.5.2", default-features = false }
 
+[dev-dependencies.objc2-foundation]
+path = "../../framework-crates/objc2-foundation"
+version = "0.2.2"
```
madsmtm, line 64: Remove the version here, otherwise we'll get cyclic dependency errors

madsmtm (changes requested):
I haven't reviewed all of it yet (use GitHub's UI to request a review when you want me to do so), only saw that you're bumping MSRV, please don't do that.

I _think_ all the examples and tests can use `c"string"` without affecting MSRV, but in the places where the actual code is, you'll have to use `unsafe { CStr::from_bytes_with_nul_unchecked(b"string\0") }` instead.

PaulDance:
> I haven't reviewed all of it yet (use GitHub's UI to request a review when you want me to do so)

Yes, sorry for the notifications, I just wanted to make things work and then check that I haven't missed anything, such as adding new tests, hence why I left is as a draft.

> saw that you're bumping MSRV, please don't do that

That was also part of the things I wanted to do before formally asking for a review, sorry. :sweat_smile:

[`CStr`](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html) was stabilized in 1.64 and the current MSRV is 1.60, so in any case it has to be bumped, right? Regarding why 1.79 specifically, which I understand constitutes a bit of an aggressive jump, it is for inline `const`s that I used to guarantee the `const fn` calls I made were indeed promoted, especially considering things like https://github.com/rust-lang/rust/pull/121557/. If you see any other way to do this without loosing these guarantees, then yes, I'll lower it as much as possible.

madsmtm:
> [CStr](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html) was stabilized in 1.64

Preeetty sure that's only [`CStr` in `core`](https://github.com/rust-lang/rust/issues/98314), the docs are just bad at visualizing this.

> guarantee the `const fn` calls I made were indeed promoted

Perhaps we can just avoid using the `const fn`s here? From my reading of the code, you seem to be using it mostly to have the definition of e.g. `BlockDescriptorCopyDisposeSignature` next to `BlockDescriptorCopyDispose` (which is commendable, but also perhaps a bit unnecessary here).

PaulDance:
> Preeetty sure that was only https://github.com/rust-lang/rust/issues/98314, the docs are just bad at visualizing this. CStr has been available since [basically forever](https://doc.rust-lang.org/1.0.0/std/ffi/struct.CStr.html).

Woops, sorry. That seems right, yes. Looking at the methods I use, I should be able to avoid any bump here, hopefully.

> Perhaps we can just avoid using the `const fn`s here?

Probably, yes. I just wanted to avoid the "let's hope promotion works similarly in previous and future compiler versions" kind of stance, although it should be acceptable :crossed_fingers:

> If not, then there are _ways_ to do the equivalent of inline `const` with traits and associated constants.

I'll try the above first and then this if it fails somehow. Thanks for the suggestion.

PaulDance:
Last thing I need to do before declaring this ready to review: try to add the `cfg(debug_assertions)`-gated encoding string safety checks.

PaulDance:
Alright @madsmtm, this should now be ready for a proper first review cycle! Sorry for the many CI runs, though, I just struggled endlessly with minor details.

crates/block2/src/traits.rs
```diff
@@ -135,3 +138,225 @@ impl_traits!(t0: T0, t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6, t7: T7, t8:
...
+///
+/// [Apple documentation]: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html
+/// [`Encoding`]: objc2::encode::Encoding
+/// [GCC documentation]: https://gcc.gnu.org/onlinedocs/gcc/Type-encoding.html
+/// [GCC method signatures]: https://gcc.gnu.org/onlinedocs/gcc/Method-signatures.html
+pub unsafe trait ManualBlockEncoding {
+    /// The function's input argument types.
+    type Arguments: EncodeArguments;
+    /// The function's return type.
+    type Return: EncodeReturn;
```
madsmtm, line 221: Did you consider not having these two parameters here? I can see how they're somewhat beneficial for ensuring that you don't accidentally pass one block encoding helper to another place where it wasn't intended, but it also makes things more cumbersome for the user than strictly necessary (and we already have the debug assertion, so it shouldn't be that big of an issue).

crates/objc2-encode/src/encoding.rs
```diff
@@ -248,6 +251,39 @@ impl Encoding {
...
+
+    /// Computes the raw signature string of the object corresponding to the
+    /// block taking `args` as inputs and returning `ret`.
+    ///
+    /// See [`objc2::encode::block_signature_string`] for more details. Prefer
+    /// using it whenever possible.
+    ///
+    /// [`objc2::encode::block_signature_string`]: https://docs.rs/objc2/latest/objc2/encode/fn.block_signature_string.html
+    pub fn block_signature_string(args: &[Self], ret: &Self) -> String {
+        // TODO: alignment?
```
madsmtm, line 269: Raising this as a comment on the PR for visibility, I'm not sure what to do about alignment myself.

madsmtm (changes requested):
Sorry for the delay, I've been away on holidays. I like the overall approach of this, and I really appreciate the thorough documentation you've put into it!

> Sorry for the many CI runs, though, I just struggled endlessly with minor details.

Don't worry, I don't pay for any of it - and it's mostly my own fault, I haven't really provided an easy way to test everything locally!

crates/block2/src/traits.rs
PaulDance: Yes, I did, but I simply followed [your comment just above](https://github.com/madsmtm/objc2/pull/636#issuecomment-2227388348) as I thought it meant you wanted it this way specifically in order to force users to explicitly think about what they are typing.

crates/tests/src/block.rs
```diff
@@ -1,8 +1,9 @@
 use core::cell::RefCell;
```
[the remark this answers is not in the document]
PaulDance: Yes, they kind of are, but I simply wanted to put the effort in reproducing a maximum of them in order to show things indeed worked as expected and that no surprising difference with the normal API would occur. I can remove them if you want though. That can also be done once the normal API is upgraded to compute the string automatically.

crates/objc2-encode/src/helper.rs
```diff
@@ -171,6 +172,26 @@ impl Primitive {
...
     }
+
+    pub(crate) const fn size(self) -> usize {
+        match self {
+            // Numbers.
+            Self::Char | Self::UChar => 1,
+            Self::Short | Self::UShort => 2,
+            Self::Int | Self::UInt | Self::Float | Self::Bool => 4,
+            Self::Long | Self::ULong | Self::Double => 8,
+            Self::LongLong | Self::ULongLong | Self::LongDouble => 16,
```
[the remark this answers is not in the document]
PaulDance: Yes, most probably. I've done all of this in a somewhat optimistic way. If you have a bit of documentation somewhere about it, it would be nice, or does it follow standard C in this regard?

crates/objc2-encode/src/encoding.rs
```diff
@@ -248,6 +251,39 @@ impl Encoding {
     pub fn equivalent_to_box(&self, other: &EncodingBox) -> bool {
         compare_encodings(self, other, NestingLevel::new(), false)
     }
+
+    /// Computes the theoretical size in bytes of the represented value type,
+    /// considering everything is packed, i.e. no alignment is computed.
+    pub fn size(&self) -> usize {
```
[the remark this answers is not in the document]
PaulDance: > This can fail to compute, so it should return an `Option<usize>`

You mean instead of returning `0` as a default? Then yes, I'll probably do that. The problem is in the use site: should we make the encoding check fail in this case, always succeed thus ignoring the case, or have a more explicit `.unwrap_or_default()` somewhere in between the two?

crates/objc2-encode/src/helper.rs
```diff
@@ -285,6 +306,37 @@ impl<E: EncodingType> Helper<'_, E> {
         }
         Ok(())
     }
+
+    pub(crate) fn size(&self, level: NestingLevel) -> usize {
```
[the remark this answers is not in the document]
PaulDance: I'm not sure it could easily since the `helper` method is from the `EncodingType` trait that `Helper` stores instances of and therefore cannot be `const`.

crates/objc2-encode/src/encoding.rs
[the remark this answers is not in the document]
madsmtm: I'd prefer `.unwrap_or(0)` at the use, rather than returning some incorrect value.

crates/objc2-encode/src/helper.rs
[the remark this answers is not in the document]
madsmtm: I think the best way would probably be something like:
```rust
Self::Char => core::mem::size_of::<core::ffi::c_char>(),
// ...
```

Though that conflicts with our MSRV of 1.60... I think we'll have to bump MSRV to 1.64 for this, though I'm probably fine with that (but it should happen in a separate PR, I'll try to get to doing that soon-ish).

crates/tests/src/block.rs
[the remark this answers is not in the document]
madsmtm: I'd prefer a few smoke tests, or if not that, then that the test file is made generic over the method (somehow) (because this is a lot of tests to maintain).

crates/block2/src/traits.rs
madsmtm: Heh, I just sorta sketched it out back then. I think now that I see it that I have a slight preference towards not having these, but no strong preference, I'll leave the decision to you.

PaulDance: I see what you mean. However, having these dummy types did help me even when simply testing: copy-pasting without appropriately changing the types made the compilation failed and therefore made me realize "oh yes, I forgot to update this". Also, when refactoring, it is easy to change the input or output type of a closure without it being written explicitly, so having a compile-time check helps in this regard as well.

All in all, I think it wasn't such a bad idea from the start if unsafety can truly arise from specifying a wrong signature string. Testing myself with the `NetworkExtension`'s only part I found requiring a signed block, but specifying a completely wrong string made no difference whatsoever if I remember correctly, so it might not be that important in the end. If safety looks to be more of a guess than a true requirement, then I'll remove the types; otherwise, I feel like keeping them is appropriate, but I might have to take a look at some examples again to see the difference.

madsmtm: > but specifying a completely wrong string made no difference whatsoever if I remember correctly, so it might not be that important in the end

See https://github.com/madsmtm/objc2/issues/442#issuecomment-2284932726 - I can imagine that specifying a random string in certain cases doesn't make a difference if you have a block with only one argument (since it's most likely going to place things correctly then), but for multi-argument blocks you're likely to run into UB. In any case, since it's invoked via. `NSInvocation`, it's definitely required for soundness.

(I posted it on the issue because I might want to reference the information later on)

madsmtm:
Hmm, I think perhaps a way to get this PR merged quicker would be to rip out the `debug_assertions` check for now (it's not robust enough for me to want to prevent users from using the API without it, and there's a bunch of issues with MSRV and such).

crates/block2/src/traits.rs
PaulDance: Ah, thanks for all the details! It might even inspire me to add more to the added safety comments in order to justify the unsafety and relieve users of potential doubts similar to what I had.

Regarding the question of whether to keep the dummy generic types associated to the smuggling trait, I've thought about it more since then and found that it could serve as self-documentation and help for readability in addition to what I had already mentioned: reading the signature string is pretty much impossible when not initiated; having the types explicitly written should help make things clearer and therefore avoid oversights and general fat-finger-ness; also, since this API is already surprising to use due to the smuggling trait, I think adding this much on top should be OK.

If it's alright with you, I think I'll prefer keeping it like this. If I haven't already, I think I'll add a mention of this in the documentation somewhere so users may understand better why something that looks useless at first glance is required.

PaulDance:
> rip out the `debug_assertions` check for now

Understandable, I can do that if you really want to, although the original idea is still interesting. If so, should I remove everything related to it, or just the checks inserted in the new block APIs and leave the rest as-is so you may work on it again in the future?

madsmtm:
> should I remove everything related to it

If you want, you can probably keep some of it, but as private, `#[allow(dead_code)]`.

PaulDance:
(As a small organization note: all my `objc2` stuff is on my work computer, so I'll be coming back to this in a bit more than two weeks after my own vacations)

PaulDance:
@madsmtm All the items should be addressed now. The only remaining issue is the MSRV bump requirement.

madsmtm:
Thanks! I'll get back to this, and will fix any nits I have myself, after I've done the MSRV bump.

In the odd case that this isn't merged in a week, ping me again ;)

PaulDance:
Thanks for the MSRV bump. It seems to work without my having to change anything, except for the tests where I use the `c""` C string literal syntax. If I remember correctly, it previously worked. Was an `--all-targets` or `--all-features` flag of sorts added to the `cargo check` used in the MSRV CI step? If so, was it intentional, i.e. should I update my code?

madsmtm:
> Thanks for the MSRV bump. It seems to work without my having to change anything, except for the tests where I use the `c""` C string literal syntax. If I remember correctly, it previously worked. Was an `--all-targets` or `--all-features` flag of sorts added to the `cargo check` used in the MSRV CI step? If so, was it intentional, i.e. should I update my code?

I think it's because even though it's guarded behind `#[cfg(test)]`, Rust still needs to know that it's valid syntax.

You _could_ put it in a separate file and import it with `#[cfg(test)] mod tests;` instead if really necessary, though I'd prefer to just not use `c"..."`, and use `"..."` + a later `cstr.to_str().unwrap()` where possible, and `from_bytes_with_nul(b"...\0").unwrap()` where not.

PaulDance:
Done :slightly_smiling_face:

madsmtm (approved):
I went through the whole thing now, thanks for the nice commit sequence! I ended up making `with_encoding` safe, since we can more the safety requirement to the `ManualBlockEncoding` trait.

And thanks for the excellent documentation and list of examples in `ManualBlockEncoding`, could not have written that better!

PaulDance:
Cool, thanks for the great collaboration!

crates/block2/src/stack.rs
```diff
@@ -190,6 +189,9 @@ where
     /// encoding information string themselves, thus obtaining a block
     /// containing it and working with these APIs.
     ///
+    /// You provide the encoding through the `E` type parameter, which shoul
```
PaulDance, line 192: `s/shoul/should/` :grin:

madsmtm: Ooops, fixed in 3d6ba5e246af8982e08ff0274435164756c8241e

PR #638, opened 2024-07-10
Metal: Reexpose safe wrappers for static `MTLDevice` getters

MarijnS95:
The safe wrappers for these static functions seem to have been commented out as it is really hard to extend a dynamic `trait` object with some concrete static functions (that exist regardless of a concrete type implementing the `MTLDevice` trait).  Instead, make them available as free functions prefixed with `mtl_device_*()`, to allow users to more easily read the "system default" device or get a list of all devices.

### DIFF
--- a/crates/objc2/src/rc/id.rs
+++ b/crates/objc2/src/rc/id.rs
@@ -162,7 +162,7 @@ pub type Id<T> = Retained<T>;
 
 impl<T: ?Sized> Retained<T> {
     #[inline]
-    pub(crate) unsafe fn new_nonnull(ptr: NonNull<T>) -> Self {
+    pub unsafe fn new_nonnull(ptr: NonNull<T>) -> Self {
         Self {
             ptr,
             item: PhantomData,
--- a/framework-crates/objc2-metal/src/device.rs
+++ b/framework-crates/objc2-metal/src/device.rs
@@ -1,16 +1,12 @@
-// use crate::MTLDevice;
-//
-// pub trait MTLDeviceExtension {
-//     pub fn system_default() -> Option<Self> {
-//         MTLCreateSystemDefaultDevice()
-//     }
-//
-//     pub fn all() -> Retained<NSArray<Self>> {
-//         #[cfg(target_os = "macos")]
-//         MTLCopyAllDevices()
-//         #[cfg(not(target_os = "macos"))]
-//         NSArray::from(MTLCreateSystemDefaultDevice())
-//     }
-// }
-//
-// impl<P: MTLDevice> MTLDeviceExtension for P {}
+use objc2::{rc::Id, runtime::ProtocolObject};
+use objc2_foundation::NSArray;
+
+use crate::MTLDevice;
+
+pub fn MTLCreateSystemDefaultDevice() -> Option<Id<ProtocolObject<dyn MTLDevice>>> {
+    unsafe { Id::from_raw(crate::generated::MTLCreateSystemDefaultDevice()) }
+}
+
+pub fn MTLCopyAllDevices() -> Id<NSArray<ProtocolObject<dyn MTLDevice>>> {
+    unsafe { Id::new_nonnull(crate::generated::MTLCopyAllDevices()) }
+}
--- a/framework-crates/objc2-metal/examples/triangle.rs
+++ b/framework-crates/objc2-metal/examples/triangle.rs
@@ -106,10 +106,7 @@ declare_class!(
             };
 
             // get the default device
-            let device = {
-                let ptr = unsafe { MTLCreateSystemDefaultDevice() };
-                unsafe { Retained::retain(ptr) }.expect("Failed to get default system device.")
-            };
+            let device = MTLCreateSystemDefaultDevice().expect("Failed to get default system device.");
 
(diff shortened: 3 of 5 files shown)


framework-crates/objc2-metal/src/device.rs
```diff
@@ -1,16 +1,24 @@
...
+//     fn all() -> Id<NSArray<ProtocolObject<dyn MTLDevice>>> {
+//         unsafe { Id::new_nonnull(crate::MTLCopyAllDevices()) }
 //     }
 // }
-//
-// impl<P: MTLDevice> MTLDeviceExtension for P {}
+
+// impl<P: MTLDevice> MTLDeviceExt for P {}
+
+pub fn mtl_device_system_default() -> Option<Id<ProtocolObject<dyn MTLDevice>>> {
```
madsmtm, line 18: I would rather have these be something like:
```rust
pub fn MTLCreateSystemDefaultDevice() -> Option<Retained<ProtocolObject<dyn MTLDevice>>> {
    extern "C" {
        pub fn MTLCreateSystemDefaultDevice() -> *mut ProtocolObject<dyn MTLDevice>;
    }
    unsafe { Retained::from_raw(MTLCreateSystemDefaultDevice()) }
}
```

That is, change the signature of `MTLCreateSystemDefaultDevice` to do the `Retained::from_raw` step internally.

crates/objc2/src/rc/id.rs
```diff
@@ -162,7 +162,7 @@ pub type Id<T> = Retained<T>;
 
 impl<T: ?Sized> Retained<T> {
     #[inline]
-    pub(crate) unsafe fn new_nonnull(ptr: NonNull<T>) -> Self {
+    pub unsafe fn new_nonnull(ptr: NonNull<T>) -> Self {
```
madsmtm, line 165: I don't want to expose this, users should use `Retained::from_raw(ptr.as_ptr()).unwrap()` instead as the unwrap should be optimized away (and if it isn't, we have `Option::unwrap_unchecked` for that).

madsmtm (changes requested):
Thanks! We _should_ work on fixing this in the generator itself, instead of fixing it locally like this, but for now I'm fine with fixing this locally for Metal.

framework-crates/objc2-metal/src/device.rs
MarijnS95: That would be lovely! I take it from this that we should keep the name to match Objective-C with CamelCase too?

Let me know when you can set that up in the generator, which would save us the hassle of reimplementing. But if you don't get to that in time, landing this now should make it easier to notice that there's some files that can be removed in the future.

crates/objc2/src/rc/id.rs
MarijnS95: Hmm that seems a bit silly when there is an API primitive available that is nicer and easier to read?

`from_raw()` uses this under the hood after all, and the only difference I see is the extra `+ Message` bound to get access to it.

madsmtm: Agree that it is a bit silly; but I'm trying to follow `std` here, and the prime examples of smart pointers, `Box` and `Arc`, only expose [`Box::from_raw`](https://doc.rust-lang.org/nightly/std/boxed/struct.Box.html#method.from_raw) and [`Arc::from_raw`](https://doc.rust-lang.org/nightly/std/sync/struct.Arc.html#method.from_raw), and I don't really want to differ from these here if I can avoid it.

framework-crates/objc2-metal/src/device.rs
madsmtm: Hmm, I've been (and will be) away for various summer holiday activities in recent times, so I'm a bit out of the loop myself. I think it would be fairly easy to implement generally, but I probably won't have time to do it before August.

I'm fine with pulling this change _now_, but in any case, I'm probably not going to cut a new release before it's implemented generally, so you could also just wait for that.

MarijnS95: I don't absolutely need these functions right now (would be nice to start using them in the example in https://github.com/Traverse-Research/gpu-allocator/pull/225).

Shall I leave this PR open then to not forget to remove this custom file when you've resolved #639? Thanks in advance!

crates/objc2/src/rc/id.rs
MarijnS95: I hadn't thought about matching `std` at all, where I don't use these functions very often (they're after all only to for raw pointers that originate from Rust at some point, i.e. temporarily creating a "user pointer" for FFI).

Here in `objc2` it seems to make sense given that your bindings already generate a `NonNull` annotation where applicable.

madsmtm:
I've fixed this and https://github.com/madsmtm/objc2/issues/639 in bae00d982d974244c40d6306497e1d9bd17d15de.

MarijnS95:
Thanks @madsmtm! The only thing I seem to be missing is turning the mention of `MTLCreateSystemDefaultDevice()` into an intradoc link inside `objc2-metal`'s module-level documentation.

madsmtm:
Good point, done in 839be2e38286904741b585150258fc460142d29b

PR #661, opened 2024-10-03
Implement runtime availability checking

madsmtm:
Add macro for runtime availability checking, with static assertions against the deployment target that allows eliding the runtime checks when unnecessary.

Part of https://github.com/madsmtm/objc2/issues/266.

Usage:
```rust
if available!(macos = 15.0, ios = 18.0, tvos = 18.0, watchos = 11.0, visonos = 2.0) {
    // Use API added in Xcode 16.0
}

// Or

#[cfg(target_os = "macos")]
if available!(macos = 15.0) {
    // Use API that's only relevant on macOS
}

// Or

if available!(macos = 10.7, ..) { // Note the `..`
    // Use API introduced in macOS 10.7, and take this branch on all other platforms too
}
```

Runtime lookup implementation tested on:
- macOS 14.7, M1 Macbook
  - `aarch64-apple-darwin`
  - `aarch64-apple-ios-macabi`
  - `x86_64-apple-ios-macabi`
  - `aarch64-apple-ios-sim`
- macOS 10.14.6, Intel Macbook
  - `x86_64-apple-darwin`
  - `i686-apple-darwin`
  - `x86_64-apple-ios`
- iOS 9.3.6, 1st generation iPad Mini
  - `armv7-apple-ios`

TODO:
- [x] Implement runtime lookup.
- [x] Optimize runtime lookup with `sysctl`.
- [x] Decide on the naming and syntax of the macro. Alternatives that I can think of are:
  - `os_version!(macos = 15.0, ..)`
  - `version!(macos = 15.0, ..)`
  - `os_available!(macos = 15.0, ..)`
  - `available!(macos(15.0), ..)`
  - `available!(macos 15.0, ..)`
  - `available!(macOS = 15.0, ..)`
  - `available!(macos = 15.0, *)`
  - `available!(macos = 15.0, _)`
- [x] ~Use macro in examples.~ No current examples where it makes sense.
- [x] Decide if a compilation error when using `available!(macos = 15.0)` on e.g. iOS is desirable, or if we just want to return `false` here?
- [x] Document macro.
- [x] Optimize runtime lookup with atomics.

### DIFF
--- a/crates/objc2/src/__macro_helpers/os_version.rs
+++ b/crates/objc2/src/__macro_helpers/os_version.rs
@@ -1,0 +1,405 @@
+//! Utilities for checking the runtime availability of APIs.
+//!
+//! TODO: Upstream some of this to `std`?
+use core::cmp::Ordering;
+use core::fmt;
+
+#[cfg(target_vendor = "apple")]
+mod apple;
+
+/// The size of the fields here are limited by Mach-O's `LC_BUILD_VERSION`.
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct OSVersion {
+    // Shuffle the versions around a little so that OSVersion has the same bit
+    // representation as the `u32` returned from `to_u32`, allowing
+    // comparisons to compile down to just between two `u32`s.
+    #[cfg(target_endian = "little")]
+    pub patch: u8,
+    #[cfg(target_endian = "little")]
+    pub minor: u8,
+    #[cfg(target_endian = "little")]
+    pub major: u16,
+
+    #[cfg(target_endian = "big")]
+    pub major: u16,
+    #[cfg(target_endian = "big")]
+    pub minor: u8,
+    #[cfg(target_endian = "big")]
+    pub patch: u8,
+}
+
+#[track_caller]
+const fn parse_usize(mut bytes: &[u8]) -> (usize, &[u8]) {
+    // Ensure we have at least one digit (that is not just a period).
+    let mut ret: usize = if let Some((&ascii, rest)) = bytes.split_first() {
+        bytes = rest;
+
+        match ascii {
+            b'0'..=b'9' => (ascii - b'0') as usize,
+            _ => panic!("found invalid digit when parsing version"),
+        }
+    } else {
+        panic!("found empty version number part")
+    };
+
+    // Parse the remaining digits.
+    while let Some((&ascii, rest)) = bytes.split_first() {
+        let digit = match ascii {
+            b'0'..=b'9' => ascii - b'0',
+            _ => break,
+        };
+
+        bytes = rest;
+
+        // This handles leading zeroes as well.
+        match ret.checked_mul(10) {
+            Some(val) => match val.checked_add(digit as _) {
+                Some(val) => ret = val,
+                None => panic!("version is too large"),
+            },
+            None => panic!("version is too large"),
+        };
+    }
+
+    (ret, bytes)
+}
+
+impl OSVersion {
+    const MIN: Self = Self {
+        major: 0,
+        minor: 0,
+        patch: 0,
+    };
+
+    const MAX: Self = Self {
+        major: u16::MAX,
+        minor: u8::MAX,
+        patch: u8::MAX,
+    };
+
+    /// Parse the version from a string at `const` time.
+    #[track_caller]
+    pub const fn from_str(version: &str) -> Self {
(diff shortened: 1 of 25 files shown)


crates/objc2/src/__macro_helpers/availability.rs
```diff
@@ -0,0 +1,384 @@
...
+        // This is mostly a re-implementation of:
+        // <https://github.com/llvm/llvm-project/blob/llvmorg-19.1.1/compiler-rt/lib/builtins/os_version_check.c>
+        //
+        // Note that this doesn't work with [zippered] `dylib`s yet, though
+        // that's probably fine, `rustc` doesn't support those either.
+        // [zippered]: https://github.com/nico/hack/blob/be6e1a6885a9d5179558b37a0b4c36bec9c4d377/notes/catalyst.md#building-a-zippered-dylib
+
+        // Rust does not yet support weak externs, and besides, I suspect it'd
+        // be dangerous to use, see:
+        // - https://reviews.llvm.org/D150397
```
BlackHoleFox, line 140: Seems that Flutter has [found a workaround](https://github.com/flutter/flutter/issues/133203), and one that won't need to exist after another XCode toolchain update? As long as there's a fallback, using a better `dlsym` implementation should be fine and keep the improved perf?

madsmtm: I actually ended up looking into the implementation of `_availability_version_check`, and found out that `sysctl kern.osproductversion` (or `sysctl kern.iossupportversion` for Mac Catalyst) exists, so I've gone with trying those first instead, and only falling back to the plist on simulators and older platforms.

madsmtm:
Noting my reasoning behind the name/syntax:
1. The name "available" more closely matches the semantics, esp. at usage sites. E.g. "if available macos 15" vs "if os version macos 15". `os_version` sounds more like it'd return the actual `OSVersion` struct.
2. `macos = 15.0` looks more cfg-like, and matches general expectations of that. Compare with e.g. `cfg!(target_feature = "crt-static")`. `macos 15.0`, while terser, looks odd compared to the usual Rust parse.
3. I chose lowercase `macos` because it matches the value of `cfg!(target_os = "macos")`.
4. For the optional trailing `..` vs. `_`, I chose `..` because it generally carries the meaning of an "all wildcard", see both the [the docs on destructuring](https://doc.rust-lang.org/reference/patterns.html#destructuring) and [RFC 3681](https://github.com/rust-lang/rfcs/pull/3681).
  `*`, as Swift uses, is overloaded in Rust to mean many more things, so I think that'd be confusing.

madsmtm:
Thanks for the review before @BlackHoleFox, really useful to get some feedback - if you feel like giving it another review, then feel free to do so now, I think I'm done with everything now.

crates/objc2/src/__macro_helpers/os_version.rs
```diff
@@ -0,0 +1,433 @@
...
+            .finish()
+    }
+}
+
+/// The combined availability.
+///
+/// This generally works closely together with the `available!` macro to make
+/// syntax checking inside that easier.
+///
+/// We use `#[cfg]`s explicitly to allow the user to omit an annotation for
```
BlackHoleFox, line 199: Outdated?

BlackHoleFox (commented):
cant really speak for the macro implementation details since I'm bad at them, but this looks nice and the usability of the macro seems fine 👀 

> I will try to upstream this implementation to the standard library later 

`std` would probably want the `CoreFoundation` approach like libcompiler-rt has since it avoids needing to link and load the entirely of `libobjc` when only writing normal Rust apps. iirc you just recently removed that linkage from `std` too. but that doesn't matter for this since its literally an objective-c crate :)

madsmtm:
> cant really speak for the macro implementation details since I'm bad at them, but this looks nice and the usability of the macro seems fine 👀

Thanks for the review though!

> > I will try to upstream this implementation to the standard library later
> 
> `std` would probably want the `CoreFoundation` approach like libcompiler-rt has since it avoids needing to link and load the entirely of `libobjc` when only writing normal Rust apps. iirc you just recently removed that linkage from `std` too. but that doesn't matter for this since its literally an objective-c crate :)

Yeah, unsure exactly which approach would be best there, since the only part that's really needed in `std` is just _something_ that can read the version in `/System/Library/CoreServices/SystemVersion.plist` (and that can be gotten from either `CoreFoundation`, `Foundation`, the `plist` crate or likely just re-implemented manually).

PR #669, opened 2024-11-14
Fix Rust panics in `exception::catch`, and make it safe

madsmtm:
The behaviour of foreign exceptions unwinding into Rust is defined since https://github.com/rust-lang/rust/pull/128321 (see also the yet unmerged reference PR: https://github.com/rust-lang/reference/pull/1226).

My understanding of how the Objective-C runtime deals with exceptions leads me to believe that the reverse is also defined, i.e. that Rust exceptions unwinding into Objective-C is sound. The big hint is that Objective-C literally has [a language primitive for catching foreign exceptions](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Exceptions/Articles/Exceptions64Bit.html#//apple_ref/doc/uid/TP40009044-SW3), `@catch(...)`.

I am yet unsure about x86 macOS (old Objective-C runtime), I think they use longjumps and whatnot, will have to test that.


madsmtm:
I tested x86 macOS, Rust panics are turned into aborts (probably because in the search phase, it tries to unwind through a frame that doesn't have CFI information)

PR #670, opened 2024-11-15
Remove `no_dead_strip` marks

### LINKED ISSUE #667: Dead links when `unstable-static-class` enabled
I've tried both `unstable-static-class` and `unstable-static-class-inlined` with LTO enabled. It seems that there are some dead links to the classes. For example, when I enable the feature `NSAttributedString` in `objc2-app-kit`, `NSPresentationIntent` is also linked to the final binary even if I don't use it. Is there some way to optimize these links out?

Berrysoft:
Closes #667 

I think there won't be a person to explain why objc ABI 2 uses `no_dead_strip` in a short time. Maybe it is better to remove it first. If there were more problems, we would know why it is necessary:)

### DIFF
--- a/crates/objc2/src/macros/mod.rs
+++ b/crates/objc2/src/macros/mod.rs
@@ -439,9 +439,10 @@ macro_rules! __statics_sel {
         /// info on "life before main".
         #[cfg_attr(
             not(all(target_os = "macos", target_arch = "x86")),
-            // Clang uses `no_dead_strip` in the link section for some reason,
-            // which other tools (notably some LLVM tools) now assume is
-            // present, so we have to add it as well.
+            // Clang uses `no_dead_strip` in the link section for some unknown reason,
+            // but it makes LTO fail to trim the unused symbols.
+            // https://github.com/madsmtm/objc2/issues/667
+            // https://github.com/llvm/llvm-project/issues/114111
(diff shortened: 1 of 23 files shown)


madsmtm (commented):
I honestly still feel like I understand this too little. For example, why do the `regular` part get stripped from the final binary?

Also not too sure about removing them from the image info / module info, from what I understand image info is kinda important for letting the (dynamic? can't remember which one, if not both) linker know that it should translate Objective-C sections.

Berrysoft:
OK, you're right. I choose to remove `no_dead_strip` for class refs and sels only.

crates/objc2/src/macros/mod.rs
```diff
@@ -442,11 +442,11 @@ macro_rules! __statics_sel {
             // Clang uses `no_dead_strip` in the link section for some reason,
             // which other tools (notably some LLVM tools) now assume is
             // present, so we have to add it as well.
```
madsmtm, line 444: Nit: Update this comment to say that we _don't_ do what Clang does (and perhaps link to https://github.com/madsmtm/objc2/issues/667)

madsmtm (approved):
Still not happy that I don't know what's going on, but I guess it's fine (with nit rebased in)

madsmtm:
Thanks!

PR #675, opened 2024-12-09
Remove unsafe from some vision framework functions

jgeralnik:
This is far from comprehensive but is all of the functions that I am currently using from vision framework.

I tried following the guide in [the README](https://github.com/madsmtm/objc2/blob/master/crates/header-translator/README.md#adding-a-new-framework-crate) - none of these functions have any meaningful limitations on their parameters.

The only point that I am unsure of is the last one:

```
If the method can throw an exception if provided with invalid inputs, it is not safe. Consider declaring a helper method that checks the preconditions first!
```

I would really like to add the function `VNImageRequestHandler::performRequests_error` as a safe function as well. It can throw an exception if it gets invalid parameters, but the rust function returns a `Result<(), Retained<NSError>>` so I'm not sure why it should be considered unsafe.

Either way - I couldn't get header-translator to generate a version of the function marked safe - tried with `class.VNImageRequestHandler.methods."performRequests:error:_".unsafe = false` but it didn't work (possibly due to the underscore)

### DIFF
--- a/framework-crates/objc2-vision/translation-config.toml
+++ b/framework-crates/objc2-vision/translation-config.toml
@@ -91,3 +91,30 @@ class.VNTargetedImageRequest.methods."initWithTargetedImageData:orientation:opti
 class.VNTargetedImageRequest.methods."initWithTargetedImageData:orientation:options:completionHandler:".skipped = true
 class.VNTargetedImageRequest.methods."initWithTargetedCMSampleBuffer:orientation:options:".skipped = true
 class.VNTargetedImageRequest.methods."initWithTargetedCMSampleBuffer:orientation:options:completionHandler:".skipped = true
+
+
+###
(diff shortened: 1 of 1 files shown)


jgeralnik:
Thanks, added VNImageRequestHandler::perform as a safe function as well

madsmtm (approved):
Cool! Could you rebase (both this PR and the `objc2-generated` branch)? Then I'll merge it after

jgeralnik:
Rebased (objc2-generated didn't need rebasing since there haven't been any changes there recently)

madsmtm:
Sorry, I should've said "squash", my bad 🙄

jgeralnik:
Ah, whoops. Squashed now

madsmtm:
Thanks!

PR #690, opened 2025-01-06
fix(dispatch2): Add `'static` bound to `exec_async`

### LINKED ISSUE #689: `Queue::exec_async` can be mis-used with just safe code
If the closure passed to `exec_async` contains a reference, it might be executed even if the referenced value is gone / has changed. It's also possible to end up with +1 `&mut T`.

Wonder if there's something that could be done, like a `'static` bound or something similar?

Example:
```rust
use core_foundation::runloop::CFRunLoopRun;
use dispatch2::Queue;
use std::thread;

struct X {
    y: u8,
}

impl Drop for X {
    fn drop(&mut self) {
        println!("dropping X {{ {} }}", self.y);
    }
}

fn mutate(x: &mut X) {
    let th = thread::current();
    let th = th.name();
    println!("mutating x ({}) from {th:?}", x.y);
    x.y += 1;
    println!(
        "x mutated (to ({})) from {th:?}, releasing exclusive access",
        x.y
    );
}

fn main() {
    thread::spawn(|| {
        let mut x = X { y: 1 };
        let queue = Queue::main();
        queue.exec_async(|| {
            mutate(&mut x);
        });
        mutate(&mut x);
    });
    unsafe { CFRunLoopRun() };
}
```

Output example (it might segfault instead):
Screenshot 2025-01-06 at 9 37 30 a m

jesdazrez:
Closes #689.

### DIFF
--- a/crates/dispatch2/CHANGELOG.md
+++ b/crates/dispatch2/CHANGELOG.md
@@ -59,6 +59,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - **BREAKING**: Use `extern "C-unwind"` instead of `extern "C"` in certain functions that required that.
 - **BREAKING**: Use `isize` instead of `usize` in certain functions where that is more correct.
+- **BREAKING**: An `F: 'static` bound was added to the following methods to make
+                sure any referenced values passed to them are borrowed for long
+                enough since they perform their `work` asynchronously (#689):
+  - `Group::exec_async`
+  - `Queue::exec_async`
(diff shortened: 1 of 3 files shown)


madsmtm (changes requested):
Could you do the same for `Group::exec_async`, `Queue::barrier_async` and `Queue::barrier_async_and_wait`?

And add a changelog entry?

jesdazrez:
sure thing!

jesdazrez:
done!

madsmtm (approved):
Thanks!

PR #698, opened 2025-01-18
Add more ServiceManagement functions

localcc:
I have allowed the `ServiceManagement` framework crate to have more functions, in particular I was missing `SMJobBless`. I see that previously functions referring to CoreFoundation were not generated. I might not have all the context on why that was, so please tell me if I did something incorrectly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1422,7 +1422,9 @@ version = "0.2.2"
 dependencies = [
  "block2",
  "objc2",
+ "objc2-core-foundation",
  "objc2-foundation",
+ "objc2-security",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


madsmtm (approved):
Thanks! This indeed makes sense to have, as `SMAppService` is only available in newer macOS versions, it was just an oversight when I added CoreFoundation support to not add this too.

I've merged this in 6b8a6d14b3fb9effd4f6fcdbdfbe74cf03fcf20d, I decided to make `objc2-core-foundation` optional as well, and to not include the changelog entry, since `objc2-core-foundation` was added in this (yet to be released) version, and there's a lot of similar additions that also aren't noted.

Issue #697, opened 2025-01-11
Any plan to publish objc2-av-foundation?


nikhilunni:
Noticing on crates.io the crate is reserved but at version `0.0.0`

madsmtm:
Yeah, I plan to do it pretty soon, check out [the date in the milestone](https://github.com/madsmtm/objc2/milestone/6).

bvsn:
Looks like no updates for AVFoundation. Any chance to have `AVAssetWriter` and `AVAssetWriterInput`? Thank you.

madsmtm:
It's coming, see https://github.com/madsmtm/objc2/pull/700

PR #700, opened 2025-01-22
New versions

madsmtm:
`objc2 v0.6.0`, `block2 v0.6.0`, `dispatch2 v0.2.0` and framework crates `v0.3.0`.

Checklist:
- [x] The branch is named `ci-full`, such that the full CI will run.
    - `objc2-proc-macros`
    - `objc2-exception-helper`
    - `objc2-encode`
    - `objc2`
    - `block2`
    - `dispatch2`
    - Framework crates
- [x] Tests have been run on macOS 10.12.
- [x] Tests have been run on macOS 10.12 32bit.

Post merge:
- [ ] Run `cargo release`.
- [ ] Create a tag on the merge commit for each new version.

Note: The framework static tests don't pass, but that only really affects CoreMIDI and DiskArbitration, so I won't block on that.


madsmtm:
Update: I've run into an issue with `objc2-app-kit` now having more than the allowed (300) number of features, I'll request an increase with the crates.io support team, and continue the publish later.

madsmtm:
Update: I got a response, discussing further with the crates.io team.

madsmtm:
Update: I have sent another email asking for a bump to just 310 features (I was asking for a lot more initially, so that I wouldn't have to ask again), since I've now received reports from two users that are confused by the half-published set of crates (https://github.com/madsmtm/objc2/issues/697#issuecomment-2612383160, and one in DM).

PaulDance:
Great! I'll try upgrading our codebase to it on Monday and will keep you informed in more details then.

PaulDance:
I've transitioned everything without too much trouble and just tested my in-development build: both the system and network extension parts still work fine.

In my case, the required refactors were not too big in the end. There were mostly about renaming and moving things around, basically. Some were more constraining, but still easy to circumvent: the types changed a bit in Foundation, for example `NSString` is not `Send` nor `Sync` anymore (probably for the better) and `NSArray` yields elements in its iterator by `Retained<_>` instead of by reference. Some were just really nice bonuses: defined classes being `Send` and `Sync` when applicable, the `msg_send_id!` -> `msg_send!` merge, `Retained::downcast` and `AnyObject::downcast_ref`, ...

Overall, a pretty good version I would say. Thanks a bunch!

Issue #699, opened 2025-01-21
Add support for MTLDevice.newLibraryWithData:error


mariusdkm:
Would it be possible to get `MTLDevice.newLibraryWithData:error` implemented?
https://developer.apple.com/documentation/metal/mtldevice/makelibrary(data:)?language=objc

It would be highly beneficial to implement `MTLDevice.newLibraryWithData:error` as it allows embedding compiled Metal library code directly into the binary. This eliminates the need to load shaders from separate files, saving time and improving performance.

I understand that this requires dispatch support—how complex would this be to implement?

Thank you for considering this feature!

madsmtm:
Dispatch support is tracked in https://github.com/madsmtm/objc2/issues/77, and some work is going on in https://github.com/madsmtm/objc2/pull/681, but I've had to delay it 'till after the next release, so it'll probably be a while before we get that support.

In the meantime, you can do something like:
```rust
use dispatch2::dispatch_data_t;
use objc2::rc::Retained;
use objc2::msg_send_id;
use objc2_metal::{MTLDevice, MTLLibrary};
use objc2_foundation::NSError;

unsafe fn newLibraryWithData_error(device: &ProtocolObject<dyn MTLDevice>, data: dispatch_data_t) -> Result<Retained<ProtocolObject<dyn MTLLibrary>>, Retained<NSError>> {
    unsafe { msg_send_id![device, newLibraryWithData: data, error: _] }
}
```

madsmtm:
(and then construct the `dispatch_data_t` manually using [`dispatch_data_create`](https://docs.rs/dispatch2/0.1.0/dispatch2/ffi/fn.dispatch_data_create.html) or similar)

madsmtm:
`MTLDevice::newLibraryWithData_error` is exposed now in 91a74f332f3db66378f457ad954f26c2d72277ca, will be released in [the next version](https://github.com/madsmtm/objc2/milestone/7).

Issue #609, opened 2024-05-01
Emit simple `#define` macros in framework crates


madsmtm:
Given things like:
```objective-c
#define MTLCounterDontSample ((NSUInteger)-1)
#define NSFoundationVersionNumber10_0	397.40
```

Emit:
```rust
const MTLCounterDontSample: NSUInteger = NSUInteger::MAX;
const NSFoundationVersionNumber10_0: c_double = 397.40;
```

Difficult because we have no way of knowing their correct type.

madsmtm:
Swift [has support for this](https://github.com/swiftlang/swift/blob/swift-6.0.3-RELEASE/docs/HowSwiftImportsCAPIs.md#macros). From [Apple's documentation](https://developer.apple.com/documentation/swift/using-imported-c-macros-in-swift):

> simple, constant-like macros, declared with the #define directive

> Macros are imported when they use literals for string, floating-point, or integer values, or use operators like `+`, `-`, `>`, and `==` between literals or previously defined macros.

Might be difficult if the macros' data isn't exposed properly to us by libclang (so hard to tell if the macro is sufficiently simple).

This is especially needed for better CoreFoundation support.

PaulDance:
This could also be nice for the string constants of `Security/AuthorizationTags.h` that are required in order to use the authorization API, namely [`AuthorizationCopyRights`](https://developer.apple.com/documentation/security/authorizationcopyrights(_:_:_:_:_:)?language=objc).

PR #710, opened 2025-02-05
dispatch2: Add `fn data_create()` which copies a slice

MarijnS95:
Fixes https://github.com/madsmtm/objc2/issues/699

# TODO
- Add a variant that takes ownership over a `Box` (or `T`?) and calls its destructor (EDIT: This requires a raw `Block` pointer, which we can probably get via a `Deref` on `GlobalBlock`).
- Make regenerated Metal code import/use the correct `dispatch_data_t` type (preferably taking `&DispatchObject` and calling `.as_raw()`?).
- Ensure calling `set_finalizer()` is still a valid thing to do?
- Can the header translator parse `#define` or are we expected to redefine them ourselves?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1516,6 +1516,7 @@ version = "0.3.0"
 dependencies = [
  "bitflags",
  "block2",
+ "dispatch2",
  "objc2",
  "objc2-app-kit",
  "objc2-core-foundation",
--- a/crates/dispatch2/src/object.rs
+++ b/crates/dispatch2/src/object.rs
@@ -1,8 +1,12 @@
 //! Dispatch object definition.
 
 use alloc::boxed::Box;
+use core::{ops::Deref, ptr::NonNull};
 
 use super::{ffi::*, queue::Queue, utils::function_wrapper, QualityOfServiceClass};
+
+// TODO: Autogenerate with https://github.com/madsmtm/objc2/issues/609
+const DISPATCH_DATA_DESTRUCTOR_DEFAULT: dispatch_block_t = std::ptr::null_mut();
 
 /// Error returned by [DispatchObject::set_target_queue].
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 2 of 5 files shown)


madsmtm:
> Add a variant that takes ownership over a `Box` (or `T`?) and calls its destructor (EDIT: This requires a raw `Block` pointer, which we can probably get via a `Deref` on `GlobalBlock`).

Yeah, could be useful. A variant which takes `&'static [u8]` could also be, and so could taking `Arc<[u8]>`.

Maybe it could even be generic over `T: AsRef<[u8]> + 'static`? Though a difficulty here is in thread-safety, I'm not sure, but I think `+ Send` would also be required (maybe depending on the queue the destructor is run on)?

Hmm... It actually seems like the block passed to `dispatch_data_create` doesn't take any parameters, so it would have to keep a pointer to the original data inside the block? That seems odd? Maybe I don't understand this API enough?

> * Make regenerated Metal code import/use the correct `dispatch_data_t` type (preferably taking `&DispatchObject` and calling `.as_raw()`?).

See https://github.com/madsmtm/objc2/pull/681; I intend to start work on that soon, that should resolve this kind of issue.

> Can the header translator parse `#define` or are we expected to redefine them ourselves?

See https://github.com/madsmtm/objc2/issues/609.

MarijnS95:
> Hmm... It actually seems like the block passed to `dispatch_data_create` doesn't take any parameters, so it would have to keep a pointer to the original data inside the block? That seems odd? Maybe I don't understand this API enough?

That's actually useful for DSTs like `[u8]` which need a fat pointer with size to reconstruct the size. `move`ing a raw pointer into a new closure solves this naturally (but we do need to allocate a new block for each call, which could be circumvented if the API returned a pointer plus size).

I'll experiment with your suggestions for `AsRef<[u8]> + 'static` later, that does sound great! Let me invesigate the threading on these queues, I'm not too familiar with them yet.

> > * Make regenerated Metal code import/use the correct `dispatch_data_t` type (preferably taking `&DispatchObject` and calling `.as_raw()`?).
> 
> See #681; I intend to start work on that soon, that should resolve this kind of issue.

Cool yeah, I see that adds a `trait AsRawDispatchObject` to get the raw object pointer, but see no generator changes: are those still needed to support converting the object on the fly?

I got the `header-translator` to run and the output tries to use the low-level FFI bindings imported from `use dispatch2::*;` but simple typedefs are (obviously) not publicly reexported from the `dispatch2` root: https://github.com/MarijnS95/objc2-generated/compare/mtl-dispatch

madsmtm:
Sorry, still haven't gotten around to finishing the `dispatch2` work. It's on [the milestone](https://github.com/madsmtm/objc2/milestone/7) for the next patch release though, so hopefully I'll get it done this month.

madsmtm:
I've integrated this in b5e5403d93a4557bbcbf3ae575242df111ea4e8c, using this PR as a base to build upon. Ended up renaming the methods a bit to better fit with the rest of the `objc2` crates.

Thanks for your help Marijn!

Will be released shortly in [the next version](https://github.com/madsmtm/objc2/milestone/7).

PR #712, opened 2025-02-11
Fix(Security): Re-define AuthorizationExecuteWithPrivileges

### LINKED ISSUE #711: `objc2_security::AuthorizationExecuteWithPrivileges` has a wrong type for its `arguments` parameter
Hi again!

[`objc2_security::AuthorizationExecuteWithPrivileges`](https://docs.rs/objc2-security/latest/objc2_security/fn.AuthorizationExecuteWithPrivileges.html) is currently generated as:

```rust
extern "C-unwind" {
    #[cfg(feature = "libc")]
    #[deprecated]
    pub fn AuthorizationExecuteWithPrivileges(
        authorization: AuthorizationRef,
        path_to_tool: NonNull<c_char>,
        options: AuthorizationFlags,
        arguments: NonNull<NonNull<c_char>>,
        communications_pipe: *mut *mut libc::FILE,
    ) -> OSStatus;
}
```

while [its official documentation](https://developer.apple.com/documentation/security/1540038-authorizationexecutewithprivileg) clearly states:

> `arguments`: An argv-style vector of strings (array of null-terminated char * pointers ending with a NULL pointer) to send to the tool.

that explicitly requires at least one null pointer in the "array" as it can be seen in [this example](https://github.com/jponge/run-with-privileges-on-osx/blob/master/main.c), as otherwise it consistently fails on `errAuthorizationToolExecuteFailure = -60031, /* The specified program could not be executed. */`, meaning the current definition almost prevents any use of the function since obtaining the last element would require transmuting a null pointer to a `NonNull`, which is highly unsafe and sensitive to compiler optimizations.

Instead, the parameter should be `arguments: NonNull<AuthorizationString>` where `AuthorizationString` already is `*const c_char`: the parameter itself is most probably not nullable, but the pointed-to values definitely are. In the meantime, downstreaming the binding manually with the type fixed seems to work just fine.

Thanks in advance,
Paul.

PaulDance:
Fixes #711.


PaulDance:
@madsmtm would it be possible to review this, please? Doing so somewhat soon would help avoid having to rebase this regularly to fix conflicts since it touches the generated modules.

madsmtm (approved):
Thanks!

> @madsmtm would it be possible to review this, please? Doing so somewhat soon would help avoid having to rebase this regularly to fix conflicts since it touches the generated modules.

Yeah sorry, I forgot about this PR when I made 349dc408e4d0c81e01667cf62748da28c6f3f6d6 🙄

PaulDance:
Wow, that was fast :smile:

madsmtm:
I typed that before you wrote your comment ;)

PaulDance:
Thanks!

Do you know if it would be possible to publish a 0.3.1 patch release for `objc2-security` containing this or do you for now prefer to keep all framework crates synchronized in versioning?

madsmtm:
I'm still unsure, but I think I'd prefer to have framework crates synchronized, yeah.

madsmtm:
Have created a milestone with a reasonable deadline (which, as you know, I'm terrible at, but hopefully this one won't be as stressful).

PaulDance:
> I'm still unsure, but I think I'd prefer to have framework crates synchronized, yeah.

Given the quantity, that's definitely understandable :ok_hand:

> Have created a milestone with a reasonable deadline

Seems fair, thanks!

> (which, as you know, I'm terrible at, but hopefully this one won't be as stressful)

:sweat_smile:

Issue #723, opened 2025-03-30
AudioToolbox docs not build


extrawurst:
currently the crate has no docs, the docs building fails with:

```
[INFO] [stderr]  Documenting objc2-audio-toolbox v0.3.0 (/opt/rustwide/workdir)
[INFO] [stderr] error[E0412]: cannot find type `MIDIChannelNumber` in this scope
[INFO] [stderr]    --> src/generated/AUAudioUnit.rs:281:35
[INFO] [stderr]     |
[INFO] [stderr] 281 |     *mut block2::Block<dyn Fn(u8, MIDIChannelNumber, NonNull<MIDICIProfile>, Bool)>;
[INFO] [stderr]     |                                   ^^^^^^^^^^^^^^^^^ not found in this scope
[INFO] [stderr]     |
[INFO] [stderr] help: you might be missing a type parameter
[INFO] [stderr]     |
[INFO] [stderr] 280 | pub type AUMIDICIProfileChangedBlock<MIDIChannelNumber> =
[INFO] [stderr]     |                                     +++++++++++++++++++
[INFO] [stderr] 
[INFO] [stderr] error[E0412]: cannot find type `MIDIChannelNumber` in this scope
[INFO] [stderr]     --> src/generated/AUAudioUnit.rs:1282:22
[INFO] [stderr]      |
[INFO] [stderr] 1282 |             channel: MIDIChannelNumber,
[INFO] [stderr]      |                      ^^^^^^^^^^^^^^^^^ not found in this scope
[INFO] [stderr] 
[INFO] [stderr] error[E0412]: cannot find type `MIDIChannelNumber` in this scope
[INFO] [stderr]     --> src/generated/AUAudioUnit.rs:1304:22
[INFO] [stderr]      |
[INFO] [stderr] 1304 |             channel: MIDIChannelNumber,
[INFO] [stderr]      |                      ^^^^^^^^^^^^^^^^^ not found in this scope
[INFO] [stderr] 
[INFO] [stderr] error[E0412]: cannot find type `MIDIChannelNumber` in this scope
[INFO] [stderr]     --> src/generated/AUAudioUnit.rs:1326:22
[INFO] [stderr]      |
[INFO] [stderr] 1326 |             channel: MIDIChannelNumber,
[INFO] [stderr]      |                      ^^^^^^^^^^^^^^^^^ not found in this scope
[INFO] [stderr] 
[INFO] [stderr] For more information about this error, try `rustc --explain E0412`.
[...]

PR #724, opened 2025-03-30
enable objc2-core-midi/MIDIMessages feature

extrawurst:
trying to solve #723


extrawurst:
@madsmtm ok I just now realize that these `Cargo.toml`s are auto generated. so this will not work :(

madsmtm:
Yeah, this will need modifications in `header-translator` somehow

Issue #753, opened 2025-05-10
IOUSBHostControllerInterface missing methods and other declarations


carlossless:
Hi,

I noticed that [`IOUSBHostControllerInterface`](https://github.com/madsmtm/objc2-generated/blob/703380ccfb3056c5a03da6ee0e24d7a593da859e/IOUSBHost/IOUSBHostControllerInterface.rs) does not have neither `getPortStateMachineForCommand:error:` nor `getPortStateMachineForPort:error:` exposed.

Not exactly sure why they were skipped in generation, but they are crucial methods in this API.

carlossless:
Additionally noticed that `IOUSBHostCISetupTransferData1wIndex` and `IOUSBHostCISetupTransferData1wLength` are also missing, though their `...Phase` counterparts are both there.

carlossless:
I'm making manual adjustments in my own objc2-generated fork to address these issues and the ones reported in #754 
https://github.com/madsmtm/objc2-generated/compare/master...carlossless:objc2-generated:fix/objc2-io-usb-host?expand=1

madsmtm:
Thanks for the report! These are missing because they're skipped in:
https://github.com/madsmtm/objc2/blob/f4e8bc173a8902722a98e1653025471c3d5d1501/framework-crates/objc2-io-usb-host/translation-config.toml#L8-L10

Ideally the issues described therein should be fixed in `header-translator`, but I'd accept a PR for these as manual fixes for now, adding extra files as `objc2/framework-crates/objc2-io-usb-host/src/xyz.rs` should be enough for that.

madsmtm:
> Does header-translator currently accept an override configuration for a methods return type nullability?

No.

> If not, would that be a desired feature for it?

Yes ;)

carlossless:
@madsmtm I created a draft pr: https://github.com/madsmtm/objc2/pull/756. It mainly serves as an example and most likely has things done not in the way you would intend them be, so any early feedback would be very much appreciated!

Aside from that, could you elaborate on this comment: https://github.com/madsmtm/objc2/blob/2c3732793ac4cc413d46a62498ed0b0361a43077/framework-crates/objc2-io-usb-host/translation-config.toml#L29 ?

madsmtm:
Ah, yeah, `IORPC.h` seems to have been added in macOS 10.15, where they ripped out 32-bit support too.

[...]

PR #756, opened 2025-05-13
config option to override method argument and return type nullability

carlossless:
This PR partially resolves https://github.com/madsmtm/objc2/issues/753#issuecomment-2871949083 by adding options in translation-config.toml to override the nullability of method argument and return types (including properties)

### DIFF
--- a/crates/header-translator/src/config.rs
+++ b/crates/header-translator/src/config.rs
@@ -574,6 +574,22 @@ pub struct CategoryData {
 
 #[derive(Deserialize, Debug, Default, Clone, PartialEq, Eq)]
 #[serde(deny_unknown_fields)]
+#[serde(rename_all = "lowercase")]
+pub enum Nullability {
+    #[default]
+    Nullable,
+    NonNull,
+}
+
+#[derive(Deserialize, Debug, Default, Clone, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct TypeOverride {
+    #[serde(default)]
+    pub nullability: Option<Nullability>,
+}
+
+#[derive(Deserialize, Debug, Default, Clone, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 pub struct MethodData {
     #[serde(default)]
     pub skipped: bool,
@@ -582,6 +598,12 @@ pub struct MethodData {
     #[serde(rename = "unsafe")]
     #[serde(default)]
     pub unsafe_: Unsafe,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_argument_overrides")]
+    pub arguments: HashMap<usize, TypeOverride>,
+    #[serde(rename = "return")]
+    #[serde(default)]
+    pub return_: TypeOverride,
 }
 
 impl MethodData {
@@ -591,6 +613,8 @@ impl MethodData {
             unsafe_: self.unsafe_,
             renamed: self.renamed.or(superclass.renamed).clone(),
             skipped: self.skipped | superclass.skipped,
+            arguments: self.arguments,
+            return_: self.return_,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


crates/header-translator/src/config.rs
```diff
@@ -582,6 +606,9 @@ pub struct MethodData {
     #[serde(rename = "unsafe")]
     #[serde(default)]
     pub unsafe_: Unsafe,
+    #[serde(rename = "override")]
+    #[serde(default)]
+    pub override_: Option<MethodOverride>,
```
madsmtm, line 611: Nit: I'd avoid this struct altogether, everything in here is an "override". Just inline `pub return_: Option<ReturnOverride>,` here (or maybe `pub return_: ReturnOverride`, since it impls Default).

crates/header-translator/src/config.rs
```diff
@@ -572,6 +572,30 @@ pub struct CategoryData {
...
 }
 
+#[derive(Deserialize, Debug, Default, Clone, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "lowercase")]
+pub enum Nullability {
+    #[default]
+    Nullable,
+    NonNull,
+    Unspecified,
```
madsmtm, line 582: Nit: Wouldn't add this as a "user-facing" option

crates/header-translator/src/method.rs
```diff
@@ -535,28 +535,29 @@ impl Method {
...
+            is_pub,
+            non_isolated: modifiers.non_isolated,
+            mainthreadonly,
+            weak_property: false,
+            must_use: modifiers.must_use,
+            encoding,
+            documentation: Documentation::from_entity(&entity, context),
+        };
+
+        method.apply_config_overrides(&data);
```
madsmtm, line 558: I'd like this to happen _before_ construction of `Method`, and I'd like it to also happen when constructing properties. Maybe make `apply_config_overrides` a free-standing function instead?

madsmtm (commented):
Short review, am at RustWeek so a bit limited time. But generally looks fine

carlossless:
@madsmtm I addressed your points and extended this override to properties and method arguments. Please let me know what you think and please give me pointers on how this should be tested, thanks!

crates/header-translator/src/method.rs
```diff
@@ -909,3 +931,19 @@ pub(crate) fn handle_reserved(name: &str) -> String {
...
+    if let Some(nullability) = &or.nullability {
+        let c_nullability = match nullability {
+            config::Nullability::Nullable => clang::Nullability::Nullable,
+            config::Nullability::NonNull => clang::Nullability::NonNull,
+        };
+
+        match ty {
+            Ty::Pointer { nullability, .. } => {
+                *nullability = c_nullability;
+            }
```
madsmtm, line 945: `Ty::Sel` and `Ty::IncompleteArray` has null-ability too. Could I get you to handle that as well?

Also, it'd be nice to emit an `warning!` for "nullabilities are equal / redundantly set in `translation-config.toml`".

framework-crates/objc2-io-usb-host/translation-config.toml
```diff
@@ -5,9 +5,9 @@ custom-lib-rs = true
 macos = "10.15"
 maccatalyst = "14.0"
 
-# Return type marked as non-null, raises `unknown error result type`.
-class.IOUSBHostControllerInterface.methods."getPortStateMachineForCommand:error:".skipped = true
-class.IOUSBHostControllerInterface.methods."getPortStateMachineForPort:error:".skipped = true
+# Return type nullability is incorrectly declared in the header. It should be `nullable` instead of `nonnull`.
+class.IOUSBHostControllerInterface.methods."getPortStateMachineForCommand:error:".return.nullability = "nullable"
+class.IOUSBHostControllerInterface.methods."getPortStateMachineForPort:error:".return.nullability = "nullable"
```
madsmtm, line 10: Nit: Could I get you to re-run the translator, and commit the result? You don't have to open a PR to `objc2-generated`, as long as you've pushed it to GH then I can find it when merging.

If it's too bothersome, then I can do it for you?

madsmtm (approved):
Thanks, looks fine now, only minor comments

framework-crates/objc2-io-usb-host/translation-config.toml
carlossless: Sure, no problem, it's here https://github.com/carlossless/objc2-generated/commit/b8d328c65b921049a9faacf6b97256aad501417f

crates/header-translator/src/method.rs
carlossless: Yep, addressed :+1:

carlossless:
@madsmtm addressed your last comments. Also, let me know if you prefer to change the warning message to something else!

framework-crates/objc2-io-usb-host/translation-config.toml
madsmtm: I meant, you have to link it in the commit in this PR. I think something like `git add generated`

carlossless: Oh, I wasn't aware this could work this way. Is this what you mean?

madsmtm: Yup, exactly.

madsmtm:
Thanks! Will merge once CI finishes

PR #759, opened 2025-05-20
iokit: enable power management mod

pronebird:
This PR adds power management for IOKit.

1. Verified that `iokit_family_err!` macro works with objc:
  ```objc
  int value = kIOPSFamilyCodeFirewire;
  NSLog(@"Value: %d", value);
  ```
  and Rust test:
  ```rs
  #[test]
  fn test() {
      let value = crate::kIOPSFamilyCodeFirewire;
      std::println!("kIOPSFamilyCodeFirewire = {}", value);
  }
  ```
  Both emit `-536838144`.
1. Some of the constants from `IOPM.h` use `iokit_family_msg` macro which for some reason is not being picked up so I simply decided to skip those defines for now
1. `iokit_family_msg` and `iokit_family_err` are used in different contexts. The latter is often used in enums which default to C `int`. But `iokit_family_msg` seems to be used in contexts where it's expected to be cast to `UInt32`. Go figure..

_Generated with XCode 16.2, sorry I can't install 16.3 on macOS 14_

### Further thoughts

If we could run C preprocessor and evaluate some of the #defines, such as the ones using `iokit_family_err` then we could perhaps plug in computed values instead of re-implementing macro definitions. Unless those vary per platform, it should be alright.

### DIFF
--- a/framework-crates/objc2-io-kit/src/lib.rs
+++ b/framework-crates/objc2-io-kit/src/lib.rs
@@ -36,6 +36,11 @@ pub type IOReturn = core::ffi::c_int; // kern_return_t
 #[allow(non_upper_case_globals)]
 pub const kIOReturnSuccess: IOReturn = 0;
 
+// IOKit/IOPM.h
+/// [Apple's documentation](https://developer.apple.com/documentation/kernel/2876248-anonymous/kiopsfamilycodeunsupported/)
+#[allow(non_upper_case_globals)]
+pub const kIOPSFamilyCodeUnsupported: core::ffi::c_int = kIOReturnUnsupported as core::ffi::c_int;
+
 // MacTypes.h
 #[allow(dead_code)]
 pub(crate) type Boolean = u8;
--- a/framework-crates/objc2-io-kit/Cargo.toml
+++ b/framework-crates/objc2-io-kit/Cargo.toml
@@ -62,6 +62,7 @@ default = [
     "libc",
     "network",
     "objc2",
+    "pwr_mgt",
     "serial",
     "usb",
 ]
@@ -106,5 +107,10 @@ hidsystem = [
     "objc2-core-foundation/CFDictionary",
 ]
 network = []
+pwr_mgt = [
+    "objc2-core-foundation/CFArray",
+    "objc2-core-foundation/CFDate",
+    "objc2-core-foundation/CFDictionary",
+]
 serial = []
 usb = []
--- a/framework-crates/objc2-io-kit/src/macros.rs
+++ b/framework-crates/objc2-io-kit/src/macros.rs
@@ -44,6 +44,20 @@ macro_rules! iokit_common_msg {
     };
 }
 
+macro_rules! iokit_family_msg {
+    ($sub:expr, $message:expr) => {
+        // (sys_iokit | sub| message)
+        (err_system!(0x38) | ($sub as i32) | $message) as u32
+    };
+}
+
+macro_rules! iokit_family_err {
+    ($sub:expr, $return:expr) => {
+        // (sys_iokit | sub| return)
+        (err_system!(0x38) | ($sub as i32) | $return) as i32
+    };
+}
+
 macro_rules! iokit_usb_msg {
     ($message:expr) => {
         // (sys_iokit | sub_iokit_usb | message)
@@ -68,6 +82,8 @@ pub(crate) use err_sub;
 pub(crate) use err_system;
 pub(crate) use iokit_common_msg;
 pub(crate) use iokit_common_msg as iokit_common_err;
+pub(crate) use iokit_family_err;
+pub(crate) use iokit_family_msg;
 pub(crate) use iokit_usb_msg as iokit_usb_err;
 pub(crate) use iokit_usb_msg;
 pub(crate) use iokit_vendor_specific_msg;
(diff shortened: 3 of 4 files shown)


framework-crates/objc2-io-kit/src/lib.rs
```diff
@@ -36,6 +36,57 @@ pub type IOReturn = core::ffi::c_int; // kern_return_t
...
+#[allow(non_upper_case_globals)]
+pub const sub_iokit_firewire: u32 = err_sub!(2);
+#[allow(non_upper_case_globals)]
+pub const sub_iokit_block_storage: u32 = err_sub!(4);
+#[allow(non_upper_case_globals)]
+pub const sub_iokit_graphics: u32 = err_sub!(5);
+#[allow(non_upper_case_globals)]
+pub const sub_iokit_networking: u32 = err_sub!(6);
+#[allow(non_upper_case_globals)]
+pub const sub_iokit_bluetooth: u32 = err_sub!(8);
```
madsmtm, line 53: Use:
```toml
const.sub_iokit_common.use-value = true
const.sub_iokit_usb.use-value = true
# etc.
```

madsmtm (commented):
> Generated with XCode 16.2, sorry I can't install 16.3 on macOS 14

You can download it at:
https://developer.apple.com/download/all/?q=xcode

Place it in `/Applications/Xcode_16.3.app`, and run:
```sh
DEVELOPER_DIR=/Applications/Xcode_16.3.app/Contents/Developer cargo run --bin header-translator -- IOKit
```

framework-crates/objc2-io-kit/src/lib.rs
pronebird: It doesn't seem to generate them out of box

```
error[E0425]: cannot find value `sub_iokit_firewire` in this scope
   --> framework-crates/objc2-io-kit/src/generated/pwr_mgt.rs:719:62
    |
719 | pub const kIOPSFamilyCodeFirewire: c_int = iokit_family_err!(sub_iokit_firewire, 0);
    |                                                              ^^^^^^^^^^^^^^^^^^ not found in this scope
```

pronebird:
> > Generated with XCode 16.2, sorry I can't install 16.3 on macOS 14
> 
> You can download it at: https://developer.apple.com/download/all/?q=xcode
> 
> Place it in `/Applications/Xcode_16.3.app`, and run:
> 
> ```shell
> DEVELOPER_DIR=/Applications/Xcode_16.3.app/Contents/Developer cargo run --bin header-translator -- IOKit
> ```

I'll have a look. How do you handle the generated files since they are put in a git submodule? I suppose I cannot include them as a part of this PR.

pronebird:
> > Generated with XCode 16.2, sorry I can't install 16.3 on macOS 14
> 
> You can download it at: https://developer.apple.com/download/all/?q=xcode
> 
> Place it in `/Applications/Xcode_16.3.app`, and run:
> 
> ```shell
> DEVELOPER_DIR=/Applications/Xcode_16.3.app/Contents/Developer cargo run --bin header-translator -- IOKit
> ```

Got it. Updated bindings in here: https://github.com/madsmtm/objc2-generated/commit/108e34e06ef57480fcfcedca7f2b661623c4b717

framework-crates/objc2-io-kit/src/lib.rs
madsmtm: Ah. Try `const.sub_iokit_firewire.skipped = false` (the define is not emitted by default).

madsmtm:
> I'll have a look. How do you handle the generated files since they are put in a git submodule? I suppose I cannot include them as a part of this PR.

Do `git add generated` while having the right branch in the submodule checked out - I'll fast-forward your commit before merging.

framework-crates/objc2-io-kit/src/lib.rs
pronebird: Fixed

pronebird:
> > I'll have a look. How do you handle the generated files since they are put in a git submodule? I suppose I cannot include them as a part of this PR.
> 
> Do `git add generated` while having the right branch in the submodule checked out - I'll fast-forward your commit before merging.

Added bindings

pronebird:
`iokit_family_msg` and `iokit_family_err` are truly bizarre. One is used as u32 another as i32

pronebird:
Alright I think it's more or less complete now. I have mapped a bunch of #defines that basically are CFString typedefs:

```c
#define kIOPMAssertPreventUserIdleSystemSleep               CFSTR("PreventUserIdleSystemSleep")
```

However they don't seem to appear in the output.

madsmtm:
> `CFSTR`

Yeah, we don't currently handle those, since they're hard to correctly emit in statics (requires that we resolve https://github.com/madsmtm/objc2/issues/258).

Also, everything with the `k` prefix would have automatically been defined, see:
https://github.com/madsmtm/objc2/blob/6061ff81ec20680ca4d1abdc7b7c445ce4d180fc/crates/header-translator/src/stmt.rs#L1726-L1734

(So there's no need to add `.skipped = false`).

pronebird:
> > `CFSTR`
> 
> Yeah, we don't currently handle those, since they're hard to correctly emit in statics (requires that we resolve #258).
> 
> Also, everything with the `k` prefix would have automatically been defined, see:
> 
> https://github.com/madsmtm/objc2/blob/6061ff81ec20680ca4d1abdc7b7c445ce4d180fc/crates/header-translator/src/stmt.rs#L1726-L1734
> 
> (So there's no need to add `.skipped = false`).
> 
> Also, please rebase `objc2-generated` now that #760 is merged.

Cheers. Let me fix that.

madsmtm (approved):
Thanks, I'm content with this as-is

PR #761, opened 2025-06-07
add unstable-core-darwin-objc feature

jbatez:
Uses the `darwin_objc` (https://github.com/rust-lang/rust/issues/145496) language feature.


crates/objc2/src/macros/mod.rs
```diff
@@ -258,6 +263,21 @@ macro_rules! sel {
     };
 }
 
+#[macro_export]
+#[cfg(all(
+    not(doc),
+    feature = "unstable-core-ffi-objc",
+    not(feature = "unstable-static-sel")
+))]
+macro_rules! sel {
```
madsmtm, line 272: Nit: Could you overwrite `__sel_inner` instead? That way, the docs on `sel!` are still available.

crates/objc2/src/macros/mod.rs
```diff
@@ -258,6 +263,21 @@ macro_rules! sel {
...
+    ($($sel:tt)*) => {{
+        let ptr = $crate::__macro_helpers::core_ffi_objc_selector!($($sel)*);
+        let ptr: *const u8 = ptr.cast_const().cast();
```
madsmtm, line 275:
```suggestion
        let ptr: *const $crate::__macro_helpers::u8 = ptr.cast_const().cast();
```

crates/objc2/src/macros/mod.rs
madsmtm: It'd also be nice to have a section explaining the `unstable-core-ffi-objc` feature flag in the `sel!` macro's `# Features` section.

framework-crates/objc2-accessibility/Cargo.toml
```diff
@@ -69,6 +69,13 @@ block2 = ["dep:block2"]
 objc2-core-foundation = ["dep:objc2-core-foundation"]
 objc2-core-graphics = ["dep:objc2-core-graphics"]
 
+unstable-core-ffi-objc = [
+    "objc2-core-foundation?/unstable-core-ffi-objc",
+    "objc2-core-graphics?/unstable-core-ffi-objc",
+    "objc2-foundation/unstable-core-ffi-objc",
+    "objc2/unstable-core-ffi-objc",
```
madsmtm, line 76: I would like for the Cargo.toml features to remain "unstable" in the sense that I'm free to remove them in a patch release - but by mentioning them across crates like this, that is no longer the case. Could we remove them here, and instead require users to manually enable the unstable features for all the `objc2` crates they depend on?

I know that'd be slightly less convenient for users, but in practice, most users already directly depend on all the `objc2-*` crates in their dependency tree (e.g. if wgpu were to test it, they'd have to enable `"unstable-core-ffi-objc"` in `objc2`, `objc2-core-foundation`, `objc2-foundation`, `objc2-core-foundation`, `objc2-metal` and `objc2-quartz-core`, all of which are direct dependencies in https://github.com/gfx-rs/wgpu/pull/5641).

madsmtm (changes requested):
Thanks! I'm probably going to wait with merging this until there's at least a rust-lang/rust PR open that we can point people towards.

jbatez:
I added support for the `class!` macro as well but just realized `extern_class!` uses `__class_inner!` directly. Since `extern_class!` needs to support `#[name = "StringLiteral"]`, that complicates my `core::ffi::objc::class!` macro which expects an identifier and not a string literal.

I wonder if my `core::ffi::objc` macros should just take string literals.

jbatez:
I made `core::ffi::objc::class!` accept either an identifier or a string literal and got it working with `extern_class!`.

madsmtm:
Re identifier vs. string literal: I tend to think the `core::ffi::objc::*` macros might actually want to only accept string literals for simplicity? Because to the runtime, that is what they are, it is only incidental that Objective-C prevents you from having weird things in them like spaces.

They aren't even limited to UTF-8, so might actually make sense to force c-string literals, to prevent interior NUL bytes by construction?

jbatez:
`#[export_name = ""]` and `#[link_name = ""]` have similar design constraints, fwiw. They disallow interior NUL bytes and I image there are platforms out there that support non-Unicode symbols that Rust has no way of handling.

jbatez:
Done. They now expect string literals and are used in new implementations of `__class_inner!` and `__sel_inner!`. Since `core::ffi::objc::selector!` doesn't expect a null-terminator, I added a new `__sel_data_terminate!` macro that conditionally null-terminates. It's used in `__sel_data!` and in `common_selectors.rs`.

Though now I'm wondering if I should change `__sel_inner` to never expect null-terminated data and instead have it decide whether or not to add a null-terminator.

jbatez:
@madsmtm is there anything else you'd like to see before this gets merged?

madsmtm (commented):
Only have a few smaller things, am otherwise happy with this as-is.

If you don't get to it within the next few days, I'll probably fix it myself, I'd like to get the framework crate parts out with v0.3.2.

crates/objc2/Cargo.toml
```diff
@@ -80,6 +80,7 @@ verify = []
 #
 # Please test it, and report any issues you may find:
 # https://github.com/madsmtm/objc2/issues/new
+unstable-core-darwin-objc = []
```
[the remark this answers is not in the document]
jbatez: I prefer the augmentation approach since in most of the crates all this feature does is enable the `darwin_objc` language feature. Maybe `unstable-rustc-darwin-objc` is a better name?

I'm a little busy this weekend so feel free to take over if you want it sooner.

Issue #338, opened 2023-01-12
Handle breaking changes when upgrading the SDK version?


madsmtm:
In https://github.com/madsmtm/objc2/pull/334 there were exactly two breaking changes:
- The signature of `NSDiffableDataSourceSnapshot::appendSectionsWithIdentifiers` was changed (added correct array type).
- `ASWebAuthenticationSessionWebBrowserSessionManager::registerDefaultsForASWASInSetupAssistantIfNeeded` was removed

Which would definitely have been possible to stub out, and avoid it being a breaking change.

But do we _want_ to? Or would we expect this functionality to be so rarely used, if ever, that it goes under "acceptable breakage"?

(Not really that important to figure out yet, since `icrate` is quite far from finished. But still important to do at some point!)

madsmtm:
In https://github.com/madsmtm/objc2/pull/762, I've decided to accept a bit of breakage in rarely used crates, but have made workarounds to avoid breaking changes in e.g. `objc2-metal`.

I find it likely that's going to be my approach going forwards as well (though when a crate is popular enough to get special treatment is still a hard tradeoff).

PR #762, opened 2025-06-09
Update to Xcode 26

madsmtm:
Just announced at [WWDC25](https://developer.apple.com/videos/wwdc2025/), this PR tracks my in-progress work on updating to that.

Yes, the version is really 26, Apple is changing the versioning scheme for all their OS versions, along with Xcode, to be the same numbering.

TODO:
- Wait until the final release of Xcode 26.
- Note breaking changes in changelog.
- Parse Swift's new "sending" attribute?
- Figure out what to do about internal `_LocationEssentials` framework.
- Figure out what to do about new `UIUtilities` "subframework".


madsmtm:
I've now fixed the issues that cropped up this update, and avoided the breaking changes for popular crates like `objc2-metal` (still a bit unclear [how we should handle these](https://github.com/madsmtm/objc2/issues/338)), so this should be ready for when the final Xcode 26 release lands.

madsmtm:
Updated for Xcode 26 Release Candidate 1.

PR #763, opened 2025-06-17
feat(io-kit): generate power state

greenhat616:
~~This is a draft PR. It should be REBASED after the generated pr got merged: https://github.com/madsmtm/objc2-generated/pull/12.~~

This is a PR for enabling PowerState in IOKit.

I manually tested those on https://github.com/gety-ai/powerstate/blob/main/src/os_impl/macos.rs, which works as expected

### DIFF
--- a/framework-crates/objc2-io-kit/translation-config.toml
+++ b/framework-crates/objc2-io-kit/translation-config.toml
@@ -15,7 +15,6 @@ module.i2c.skipped = true
 module.iokitmig.skipped = true
 module.kext.skipped = true
 module.ndrvsupport.skipped = true
-module.ps.skipped = true
 module.sbp2.skipped = true
 module.scsi.skipped = true
 module.storage.skipped = true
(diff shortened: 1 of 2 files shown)


framework-crates/objc2-io-kit/translation-config.toml
```diff
@@ -15,7 +15,7 @@ module.i2c.skipped = true
 module.iokitmig.skipped = true
 module.kext.skipped = true
 module.ndrvsupport.skipped = true
-module.ps.skipped = true
+# module.ps.skipped = true
```
madsmtm, line 18: Nit: Just remove this line instead of uncommenting

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "generated"]
 	path = generated
-	url = https://github.com/madsmtm/objc2-generated.git
+	url = https://github.com/gety-ai/objc2-generated.git
```
madsmtm, line 3: It should be enough to update the submodule itself below, I think you can remove this change and CI will still pass

PR #772, opened 2025-07-23
Update to Xcode 26 Beta 4

MarijnS95:
Targets #762
See also https://github.com/madsmtm/objc2-generated/pull/13

I've been following the betas closely and need to stay on the same/latest APIs for compatibility.

Note that I didn't test if some of the `header-translator` "hacks" can already be reverted - I'm mostly working on Metal.


simlay (commented):
I think Mads is on a sailboat. I've not seen him on the internet for a few weeks. I looked at [last July and mads was similarly not around](https://github.com/madsmtm?tab=overview&from=2024-12-01&to=2024-12-31) so I'm not worried yet.

While I do have merge rights to the objc2 repo and publish rights for a bunch of the generated crates, I don't have merge/push rights on https://github.com/madsmtm/objc2-generated which is pretty needed.

PR #779, opened 2025-09-03
Fix(NetworkExtension): Make [NWHostEndpoint hostname] return-nullable

### LINKED ISSUE #778: `-[NWHostEndpoint hostname]` can panic on "unexpected NULL returned from `-[NWAddressEndpoint hostname]`"
We have observed the following panic a few times in production:

```
Got a panic: panicked at /builds/[...]/vendor/deps/objc2-network-extension-0.3.1/src/generated/mod.rs:9479:5:
unexpected NULL returned from -[NWAddressEndpoint hostname]
0: __mh_execute_header
1: __mh_execute_header
2: __mh_execute_header
3: __mh_execute_header
4: __mh_execute_header
5: __mh_execute_header
6: __mh_execute_header
7: __mh_execute_header
8: __mh_execute_header
9: __mh_execute_header
10: __mh_execute_header
11: __mh_execute_header
12: __mh_execute_header
13: __mh_execute_header
14: __mh_execute_header
15: __mh_execute_header
16: __mh_execute_header
17: __mh_execute_header
18: __mh_execute_header
19: __mh_execute_header
20: __mh_execute_header
21: __mh_execute_header
22: __mh_execute_header
23: __mh_execute_header
24: __mh_execute_header
25: __mh_execute_header
26: __mh_execute_header
27: __pthread_cond_wait
```

(super duper useful backtrace, yay)

From the two separate cases I easily have access to at the moment, both machines are under macOS 15.6.1.

We call `[NWHostEndpoint hostname]` on the endpoints given to us from the NetworkExtension `FilterDataProvider` API as part of the socket flows after downcasting them from `NWEndpoint` to `NWHostEndpoint`, so we don't really have control over the values themselves nor how their are built. This seems like a macOS or bindings issue to me.

Regarding the bindings lead, `[NWHostEndpoint hostname]` is [generated in objc2 with a non-null return type](https://github.com/madsmtm/objc2-generated/blob/v0.3.1/NetworkExtension/mod.rs#L9493-L9497) and [in the SDK, the properties also don't have the `nullable` attribute](https://github.com/roblabla/MacOSX-SDKs/blob/3e1f09a627edb40f83d000739ea9365e0104803c/MacOSX15.0.sdk/System/Library/Frameworks/NetworkExtension.framework/Versions/A/Headers/NWHostEndpoint.h#L46-L52). The SDK headers also use the `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` macros to surround the definitions, but I don't know what it really means or does. When looking at [`nw_endpoint_get_hostname`](https://github.com/roblabla/MacOSX-SDKs/blob/3e1f09a627edb40f83d000739ea9365e0104803c/MacOSX15.0.sdk/System/Library/Frameworks/Network.framework/Versions/A/Headers/endpoint.h#L106-L122) that it should only wrap if I understand correctly, it clearly states that it:

> Returns the hostname string for the endpoint, or NULL
> if the endpoint is not of type nw_endpoint_type_host.

and considering that it:

> Retrieves the hostname string for a network endpoint with
> the type nw_endpoint_type_host or nw_endpoint_type_url.

it should mean that the NULL case really is reachable, which should in turn imply that the Obj-C wrapper headers are wrong in this aspect, if not both even as I don't see `_Nullable` for `nw_endpoint_get_hostname` either, right?

If so, what do you think would be the best course of action for fixing this? Even though I don't have much hope for it, I would first try reporting this to Apple, but then it would mean skipping the class in the generation and then manually redefining it with the fixed return type?

Also, interestingly, `NWHostEndpoint` seems to have the `NWAddressEndpoint` class name at runtime for some obscure reason, because `ClassType::class` is used on the receiver when generating the panic from what I can see in the `msg_send!` implementation details. The only place I see the latter is in [`Network.tbd`](https://github.com/roblabla/MacOSX-SDKs/blob/3e1f09a627edb40f83d000739ea9365e0104803c/MacOSX15.0.sdk/System/Library/Frameworks/Network.framework/Versions/A/Network.tbd#L12651), but I don't really know what it means because [`NWHostEndpoint`](https://github.com/roblabla/MacOSX-SDKs/blob/3e1f09a627edb40f83d000739ea9365e0104803c/MacOSX15.0.sdk/System/Library/Frameworks/Network.framework/Versions/A/Network.tbd#L12654) is also there. It does mention `nw_endpoint_type_address`, but I don't know why it wouldn't be reflected in a proper class available in the SDK as per the usual 🤷

PaulDance:
Fixes #778.


madsmtm:
Thanks!

I won't merge this 'till after I begin on v0.4, so don't worry about keeping it rebased.

Issue #782, opened 2025-09-10
Tracking issue for APIs that are incorrectly marked safe


madsmtm:
There's a lot of APIs in the framework crates, and some of them are unfortunately incorrectly marked as safe (which means they're unsound), to varying degrees of severity.

I will leave comments below for each known issue, and mark them as resolved once a new release with these fixes have been released.

Feel free to comment yourself if you find an API that might be incorrectly marked safe, I would love to know about it! For other soundness issues, please [open a new issue](https://github.com/madsmtm/objc2/issues/new).

madsmtm:
[`CFPreferencesCopyMultiple`](https://developer.apple.com/documentation/corefoundation/cfpreferencescopymultiple(_:_:_:_:)?language=objc) takes an untyped `&CFArray`, it should take `&CFArray<CFString>`.

madsmtm:
Several functions in `dispatch2` takes a raw pointer `NonNull<dispatch_object_s>` (`dispatch_get_context`, `dispatch_activate`, `dispatch_suspend`, `dispatch_resume`).

madsmtm:
The methods `SCNetworkInterface::media_sub_types`, `SCNetworkInterface::media_sub_type_options` and `SCNetworkInterface::set_media_options` take an untyped `CFArray`.

madsmtm:
`NSWindow::delegate`/`NSWindow::setDelegate` may be unsound if using an `NSSavePanel`, especially since `NSOpenSavePanelDelegate` is not a sub-protocol of `NSWindowDelegate` (so you can break type-safety and convert `ProtocolObject<dyn NSOpenSavePanelDelegate>` to `ProtocolObject<dyn NSWindowDelegate>`, or vice-versa).

madsmtm:
Safety in a lot of Metal APIs may not be correctly marked, see https://hackmd.io/@andyl-moz/S1hd5V7XZl for discussion.

uqio:
`NSDictionary::to_vecs` is safe, but is unsound for any dictionary whose `count` disagrees with what `getObjects:andKeys:` actually writes.

`to_vecs_unchecked` (in `objc2-foundation/src/dictionary.rs`), which the safe `to_vecs` calls:

```rust
let len = self.len();
let mut keys = Vec::with_capacity(len);
let mut objs = Vec::with_capacity(len);
unsafe { self.getObjects_andKeys(objs.as_mut_ptr(), keys.as_mut_ptr()); }
[...]

Issue #685, opened 2024-12-21
Safety of methods appear to be illogical


kvark:
Looking at [MTLInstanceAccelerationStructureDescriptor](https://docs.rs/objc2-metal/0.2.2/objc2_metal/struct.MTLInstanceAccelerationStructureDescriptor.html#method.setInstanceCount), I see `setInstanceCount` being safe but `setInstanceDescriptorType` being `unsafe`. That doesn't make sense.

madsmtm:
Yeah, the simple reason is that [`setInstanceCount` is explicitly marked as safe](https://github.com/madsmtm/objc2/blob/d04845b57166d8d941e2492f357564219874e05c/framework-crates/objc2-metal/translation-config.toml#L72), while `setInstanceDescriptorType` just... Isn't (yet).

The issue is fundamental, it's unsound for us to mark something as safe without a human in the loop (this _is_ FFI to a C-like language, and C can do anything, regardless of the signature of the function).

That said, Objective-C _is_ more constrained, and _usually_ doesn't do crazy things, so I do have a few thoughts and ideas for automatically marking certain things as safe in certain cases - just haven't gone down that route yet, my priorities have been on making the bindings correct before making them convenient ;)

kvark:
Got it, yeah. It's unfortunate if we have to go and inspect every method for safety. Compromises the idea of auto-generated bindings to start with.
Please feel free to close if this is already tracked.

madsmtm:
I thought I'd written it down somewhere, but no, the ideas for doing some of it automatically have only existed in my head... So I'll keep this issue open, and track it here.

madsmtm:
Heads up: I implemented the "marking things safe automatically" idea primarily in https://github.com/madsmtm/objc2/pull/783, and have opened https://github.com/madsmtm/objc2/pull/792 to do the same for Metal. Review's welcome, though it's likely I'll be merging it rather quickly (I'm behind schedule on `v0.3.2`, and I'd like to get this out with that).

PR #783, opened 2025-09-10
Automatically mark `objc2-foundation` as safe

madsmtm:
I've been improving `header-translator` in [the past few weeks](https://github.com/madsmtm/objc2/commits/master/crates/header-translator?since=2025-08-21&until=2025-09-10) to allow it to have a notion of when a method/function signature is safe or unsafe. This is used to emit a relevant `# Safety` comment, see for example [`NSMenuItem::setAction`](https://github.com/madsmtm/objc2-generated/blob/1d84ccaa2dcba1fe8fd6c74d05aafa0e27572cbd/AppKit/NSMenuItem.rs#L375-L377) (the messages here aren't perfect, but they're a lot better than nothing).

There's two cases that it cannot fully handle: **bounds information** and **restrictions only present in documentation**. But apart from these, I believe that `header-translator` knows about all the other classes of memory safety issues. I've gone over this in a lot more detail in [this document](https://github.com/madsmtm/objc2/blob/66df2dd42e35623154cff2e61ec6984102137b61/crates/objc2/src/topics/frameworks_soundness.md).

This _vastly_ simplifies the safety review process, since as a reviewer, you only have to look for APIs where one of the two aforementioned issues are present. By now, I have enough confidence in this that **I propose to switch from "unsafe by default" to "safe by default"** in reviewed framework crates. This means that instead of having to manually marking each and every API as safe, we now instead mark the few APIs that do odd things as `unsafe`.

In this PR, I have done this for Foundation, see `objc2-foundation/translation-config.toml`. The `unsafe-default-safety` annotations change the default, and the rest of the annotations mark APIs as `unsafe` that I found that don't follow the normal rules. I intend to do the same for other framework crates later.

---

**This is completely at odds with how safety is normally done in Rust!**[^precedent] In Rust, we prefer `unsafe` blocks to be scoped as tightly as possible. But here, I'm effectively applying a huge `unsafe` block around the entire crate that says "all of this is safe because the signature is safe", and only selectively unmarking APIs as "wait, actually this isn't safe". A single slip-up, and boom, the library is unsound.

[^precedent]: That there is a bit of precedent in the ecosystem for this in [`autocxx::safety!`](https://docs.rs/autocxx/0.30.0/autocxx/macro.safety.html), but it is far from widespread.

Let me give my reasoning for why I propose doing it anyway:

When users use one of the framework crates, they currently have to write `unsafe` all over the place. As an example, in `winit-appkit`, there's currently 166 uses of `unsafe {`. I estimate that around half of those would disappear if Foundation and AppKit were appropriately marked safe. And this is in a library which I maintain where I've [previously marked huge swaths of these APIs as safe for this purpose](https://github.com/madsmtm/objc2/pull/475), in other projects this problem is further exacerbated.

The true danger here is not so much the annoyance of the extra annotations, but that maybe around 40 of the remaining `unsafe {` instances in `winit-appkit` are _actually_ unsafe, and require appropriate `// SAFETY` comments - but this is masked by the huge number of not-really-worth-bothering-with-`unsafe`.

Thus, to avoid our users becoming "safety blind", and no longer paying attention to bigger safety concerns like a `transmute`, a raw pointer deref, `NSWindow::setReleasedWhenClosed` or `Retained::cast_unchecked`, **we need to mark methods safe whenever possible**. There is, however, more than 6000 methods in Foundation and more than 10000 methods in AppKit, which is **far too large an amount for anyone to confidently review**.

We are in a bind! [The project's goals of soundness and not pushing safety onto users](https://github.com/madsmtm/objc2/blob/54dee0dfb91ceeae34ceea8a26de34479bdb7ebf/crates/objc2/src/topics/goals.md) are in tension, and something has to give! I believe that **endangering perfect soundness here will ultimately make user's code much safer**.

That is not to say I haven't reviewed this; I have, and I'm fairly confident that I caught the most egregious instances. But again, there is 4000 methods that I'm marking safe in this PR, there's bound to be mistakes somewhere. I have created https://github.com/madsmtm/objc2/issues/782 to track the ones that we discover.

---

CC people that might have an opinion on this: @simlay @silvanshade @not-jan @MarijnS95 @pcwalton @waywardmonkeys @extrawurst @tmandry @PaulDance @ryanmcgrath @complexspaces @pronebird @kvark @nicoburns @amodm

This is in a sense a policy change from my side, hence why I'm opening this for wider discussion. What do you think? Is this a no-go for you? Should I do it differently? Is there something that is unclear?

I'd also be curious if you know of APIs (Foundation or otherwise) that should definitely be marked `unsafe`, even though their signature might appear safe.

### DIFF
--- a/crates/objc2/src/__macros/msg_send/mod.rs
+++ b/crates/objc2/src/__macros/msg_send/mod.rs
@@ -396,10 +396,10 @@ pub use self::retained::*;
 /// use objc2_foundation::{NSNumber, NSString, NSURLComponents};
 ///
 /// let components = unsafe { NSURLComponents::new() };
-/// unsafe { components.setPort(Some(&NSNumber::new_i32(8080))) };
-/// unsafe { components.setHost(Some(&NSString::from_str("example.com"))) };
-/// unsafe { components.setScheme(Some(&NSString::from_str("http"))) };
-/// let string = unsafe { components.string() };
+/// components.setPort(Some(&NSNumber::new_i32(8080)));
+/// components.setHost(Some(&NSString::from_str("example.com")));
+/// components.setScheme(Some(&NSString::from_str("http")));
+/// let string = components.string();
 ///
 /// assert_eq!(string.unwrap().to_string(), "http://example.com:8080");
 /// ```
--- a/crates/objc2/src/topics/FRAMEWORKS_CHANGELOG.md
+++ b/crates/objc2/src/topics/FRAMEWORKS_CHANGELOG.md
@@ -24,6 +24,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - `OpenGLES` / `objc2-open-gl-es`.
   - `XCTest` / `objc2-xc-test`.
   - `XCUIAutomation` / `objc2-xc-ui-automation`.
+* Automatically marked a bunch of functions safe in:
+  - `Foundation` / `objc2-foundation`.
 
 ### Changed
 * Updated SDK from Xcode 16.3 to 16.4.
--- a/crates/objc2/src/topics/run_loop.md
+++ b/crates/objc2/src/topics/run_loop.md
@@ -15,11 +15,11 @@ In non-graphical applications, you get the thread's current `NSRunLoop`, and run
 use objc2_foundation::{NSDate, NSDefaultRunLoopMode, NSRunLoop};
 
 fn main() {
-    let run_loop = unsafe { NSRunLoop::currentRunLoop() };
+    let run_loop = NSRunLoop::currentRunLoop();
 
     // Set up timers, sources, etc.
 
-    let mut date = unsafe { NSDate::now() };
+    let mut date = NSDate::now();
     // Run for roughly 10 seconds
     for i in 0..10 {
         date = unsafe { date.dateByAddingTimeInterval(1.0) };
--- a/crates/objc2/tests/backtrace.rs
+++ b/crates/objc2/tests/backtrace.rs
@@ -12,8 +12,8 @@ fn merge_objc_symbols(exc: &NSException) -> Vec<String> {
     // demangled symbols.
     let mut demangled_symbols = vec![];
 
-    let nssymbols = unsafe { exc.callStackSymbols() };
-    let return_addrs = unsafe { exc.callStackReturnAddresses() };
+    let nssymbols = exc.callStackSymbols();
+    let return_addrs = exc.callStackReturnAddresses();
 
     for (nssymbol, addr) in nssymbols.iter().zip(return_addrs) {
         let addr = addr.as_usize() as *mut c_void;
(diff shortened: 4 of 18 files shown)


nicoburns:
Hmm... I have mixed feelings about this. I would definitely be keen to reduce "unsafe fatigue". But I do worry somewhat about it being "safe by default".

You mention some invariants are only contained in documentation, and I wonder if we could look at automatically pulling the documentation into doc comments. That might make it more obvious to people that there are invariants to uphold (when there are).

Apple seems to make their docs available in JSON format (e.g. https://developer.apple.com/tutorials/data/documentation/foundation/nsstring/linerange(for:).json?language=objc), and https://github.com/NSHipster/sosumi.ai/blob/main/src/lib/fetch.ts has more details on the URL conventions.

madsmtm:
> Apple seems to make their docs available in JSON format

Damn, I have been looking for an actual machine-readable format like that for _ages_! Maybe they changed it recently? In any case, definitely something I'll look into doing - tracking in https://github.com/madsmtm/objc2/issues/309.

mrobinson:
I agree that the pragmatic approach seems best in this case. It could be the case that a a method might be marked safe when it should be marked as unsafe with this approach, but these cases can be fixed iteratively. Additionally, the unsafe marking doesn't automatically fix any problems that the crate user has created via a misuse of the API and, as you say @madsmtm, can hide other ones.

simlay:
I think this is a pretty pragmatic plan. In my various hackery, I frequently commit a bad practices and ignore the "unsafeness" a lot when I use various objc2 generated bindings and throw a "todo" line in for future me to forget about. I know that a lot of these calls aren't all that dangerous but I feel bad about it. If all the things are unsafe and we ignore them, we're degrading the definition of `unsafe`.

Tangentially related, there's a bit of similarity between `objc2` and [jni-rs](https://github.com/jni-rs/jni-rs). Certain things I look at with jni-rs and feel less safe about (this is not a slight at rib/jni-rs, rib is awesome). If there's a typo in a [java class name string](https://github.com/jni-rs/jni-rs/blob/808f60510529b1b6eff69d2a78f2c0b1664813fb/tests/java_integers.rs#L16), it'll blow up/return an error which is basically what happens with `msg_send` (requires unsafe) in objc2. Similarly there are [some unsafe functions and sometimes why they're unsafe](https://docs.rs/jni/latest/jni/struct.JNIEnv.html#method.get_array_elements) and the arguments follow similar arguments mentioned in the [framework soundness doc](https://github.com/madsmtm/objc2/blob/66df2dd42e35623154cff2e61ec6984102137b61/crates/objc2/src/topics/frameworks_soundness.md).

Anyway, I guess I don't know how to go about validating/invalidating the safety of a given class/function/etc. Like:
```
# Foundation's properties are well-behaved.
unsafe-default-safety.property-getters = true
unsafe-default-safety.property-setters = true
```
Which I trust but how do we go about this for future frameworks? It's outside the scope of this issue and I know it was partially raised in https://github.com/madsmtm/objc2/issues/685.

madsmtm:
It seems the consensus is to go forwards with this change, so I will do so. Again, thanks all for the input, I truly appreciate it!

And of course, if you at some point find a safety mistake, please do tell me about it in https://github.com/madsmtm/objc2/issues/782.

PR #786, opened 2025-09-24
Add network framework

pronebird:
This is a boilerplate PR for adding bindings for network framework. 

### objc-network

- Add `nw_object` macros to create network (NW) objects
- Introduce `NWRetained` type. Same as `Retained` but calls into `nw_retain` and `nw_release`.
- Add placeholders for all NW types
- Add sample wrapper for `Path` and `PathMonitor` types raising some questions on how we should handle pointers, bindings.

### Changes to header-translator

- Recognize `nw_` types in order to use `NWRetained` instead of `Retained`
- Handle `NW_RETURNS_RETAINED`, `NW_NOESCAPE`, `NW_SWIFT_DISABLE_ASYNC`, `NW_ENUM`

### Questions

For what it's worth, it compiles. However I have raised some questions in the code, maybe you could help me to fix the missing bits. Of course in general I'd like to know if the direction I took is the one that you envisioned for network framework and other frameworks too.

#### Use of self in wrapper types

A lot of `nw_*` methods are generated with network objects passed as `NonNull<T>`. This is handled this way in `header-translator` and I had copied this bit verbatim assuming there is no difference with `dispatch2`. However It feels weird that i need to use `NonNull::from(self)`, i.e:

```rs
pub fn prohibit_interface_type(&mut self, interface_type: InterfaceType) {
        unsafe {
            nw_path_monitor_prohibit_interface_type(
                // fixme: should be able to use &PathMonitor?
                NonNull::from(self),
                interface_type,
            );
        }
    }
```
 
What woudl be the right approach?

#### Arrays generated as ArrayUnknownABI

A 6-char C-array was generated as `ArrayUnknownABI`. What should I specify and where to generate a normal C array? Currently skipping the symbol and providing it manually via `lib.rs`:

```
#[allow(non_camel_case_types)]
pub type nw_ethernet_address_t = [core::ffi::c_uchar; 6];

### CODE REVIEW

File: framework-crates/objc2-network/src/protocol_options.rs [hunk 1]
```diff
@@ -0,0 +1,14 @@
+nw_object!(
+    #[doc(alias = "nw_protocol_definition_t")]
+    pub struct ProtocolDefinition;
+);
```
madsmtm, line 4: I'm leaning towards modifying `header-translator` to output these instead of manually emitting them. I didn't do so for the Dispatch types, but there's ~30 types using `NW_OBJECT_DECL`, so here it might make more sense?

Besides, I am actually regretting it a bit for Dispatch, because we're missing the documentation comments for its types.

madsmtm: So that means adding it as something like `Stmt::NetworkDecl` (with a lot of fields copied from `Stmt::OpaqueDecl`).

framework-crates/objc2-network/src/path_monitor.rs
```diff
@@ -0,0 +1,76 @@
...
+);
+
+impl PathMonitor {
+    pub fn new() -> NWRetained<Self> {
+        nw_path_monitor_create()
+    }
+
+    pub fn new_with_interface_type(interface_type: InterfaceType) -> NWRetained<Self> {
+        nw_path_monitor_create_with_type(interface_type)
+    }
```
madsmtm, line 17: There should be a way for `header-translator` to automatically emit these as associated functions, you might need to modify the `find_fn_implementor` logic in `name_translation.rs` (it's kind of a mess tbh).

madsmtm (changes requested):
Thanks for doing this! I had a quick look, I'll try to give a more detailed review later, but I'm a bit busy with v0.3.2 and such atm. (yeah, sorry, this won't go in that release).

I actually did a bit of work on this in the past, have pushed https://github.com/madsmtm/objc2/tree/wip-network if you want to take a look. IIRC I got stuck on the naming of things. I think I'd like the prefix `NW`, e.g. `nw_path_t` -> `NWPath` like Swift does it, but that's already used by NetworkExtension, so we might have to do something different here.

> Use of self in wrapper types

This is handled in `global_analysis.rs`, I suspect it'll fix itself if you figure out the name translation stuff (it works in my branch above).

> Arrays generated as ArrayUnknownABI

I'd actually rather lean towards skipping `nw_ethernet_channel_send` and `nw_ethernet_channel_set_receive_handler` altogether for now, arrays do have a weird ABI in function arguments, I can't recall the details but it's something like that they actually need to be passed by reference (and I'm not sure `rustc` handles this correctly?).

> Type-aliases for generated types

No, I'd strongly prefer that, if we expose them, that type aliases like these aren't wrapped in `NWRetained` (even though that's technically incorrect, the type alias is a pointer).

pronebird:
> > I'd like the prefix `NW`, e.g. `nw_path_t` -> `NWPath` like Swift does it, but that's already used by NetworkExtension
> 
> From what I can see in [the docs](https://developer.apple.com/documentation/networkextension/nwhostendpoint?language=objc), when searching for `nw` in the bottom left bar, almost everything that comes up is either some kind of Obj-C class around a Network-like type that is now deprecated in favor of the Network type or something directly from Network itself. Furthermore, these constitute a minority compared to the rest that starts with `NE`, as one would expect more. It would seem like these wrappers should have been directly in Network, hence the naming, but were made available from NetworkExtension instead for some obscure reasons.
> 
> Considering this, I think using the `NW` prefix here would be preferable in order to remain consistent with how other frameworks have their items named and how this one already has the `nw_` prefix in most items, if it's technically feasible. It would introduce some slight confusion, but it would effectively represent what is exposed in a transparent manner, i.e. the confusion is more on them than on us. Also, there shouldn't be too much of a risk of clashing, except in downstream imports that can always be adapted, right?

I think that class prefix is only relevant for Swift/Objc due semantics of how modules work in Swift. Rust does not have this problem, hence a few deprecated types from NetworkExtension that carry the same name shouldn't matter. On the other side I don't know why the prefix is even used in Rust bindings, it looks ugly. Personally I don't care all that much for prefix, beyond all my goal is be able to use network framework from Rust :)

framework-crates/objc2-network/src/path_monitor.rs
pronebird: I am not entirely familiar with the header-translation pipeline. Currently it seems that many things are handled in different places but maybe that's because I am not familiar with how it works. 

For example:

```
nw_txt_record_t nw_txt_record_copy(nw_txt_record_t txt_record);
```

is generated into

```rs

impl NWTxtRecord {
  pub fn copy(txt_record: Option<&NWTxtRecord>) -> Option<NWRetained<NWTxtRecord>> { .. }
}
```

Given that we change implementer to `NWTxtRecord`, I'd have expect the first argument be replaced with `&self` but I am not sure where this transform should happen.

madsmtm:
I've rebased, fixed and merged this now in f12f1009a5bfb89bf37feb6583beab1e507f23a2, thanks for your work in here! (I had kinda forgotten about this PR, sorry about the delay).

I went with renaming the deprecated NetworkExtension classes from `NW*` to `NE*` (so `objc2_network_extension::NWEndpoint` -> `objc2_network_extension::NEEndpoint`). Not completely correct, but it's good enough for a deprecated API.

> On the other side I don't know why the prefix is even used in Rust bindings, it looks ugly. Personally I don't care all that much for prefix. Whichever way we go, my primary objective is to be able to use network framework from Rust :)

Idk, I've become so used to it by now. I do think it's easy to get confused by seeing `Data` and having to figure out "oh, you mean `CFData`, not `NSData`". But drop a line in https://github.com/madsmtm/objc2/issues/284 if you have a suggestion for a better design ;)

PR #790, opened 2025-09-29
Mark Metal as automatically safe

MarijnS95:
Supersedes #773 #774


madsmtm:
Sorry, I actually also worked on this just today, so we kinda raced on this. I'll close this in favour of https://github.com/madsmtm/objc2/pull/792, but thanks anyway!

PR #792, opened 2025-09-29
Automatically mark `objc2-metal` as safe

madsmtm:
https://github.com/madsmtm/objc2/pull/783 relaxed our safety policy to allow us to mark things as automatically safe. This would be very useful to do in `objc2-metal` as well, which is what this PR does.

There are a lot of wrinkles to that though, since Metal is fairly low-level, and there's a lot of seemingly innocuous things that can cause memory safety issues down the line, it effectively allows doing arbitrary things on another processor (the GPU). One example of this is `MTLFunction`, which amounts to an `unsafe extern "C" fn()` that the GPU will gladly call with whatever arguments you specify, and thus we have to mark functions taking it as unsafe.

In a similar vein, all `MTLResource`s are marked as unsafe, as accesses to these are unsynchronized, can be unretained, and need to be bound before use. `MTLBuffer` is also untyped. See also the more broad questions raised in https://github.com/madsmtm/objc2/pull/773#issuecomment-3218100934.

For now, I'd like to postpone the discussion of bounds safety, I've used `unsafe-default-safety.not-bounds-affecting = true` to do that. It still has a lot of false positives, I have fixed some of those in this PR, but we'll have to deal with the rest later. I have opened https://github.com/madsmtm/objc2/issues/793 to track this.

Replaces https://github.com/madsmtm/objc2/pull/773, https://github.com/madsmtm/objc2/pull/774 and https://github.com/madsmtm/objc2/pull/790. Closes https://github.com/madsmtm/objc2/issues/685. This is important for https://github.com/gfx-rs/wgpu/pull/5641 and for making Metal 4 nicer to use.

### DIFF
--- a/framework-crates/objc2-metal/translation-config.toml
+++ b/framework-crates/objc2-metal/translation-config.toml
@@ -154,13 +154,31 @@ class.MTLInstanceAccelerationStructureDescriptor.methods."setInstanceDescriptorB
 class.MTLInstanceAccelerationStructureDescriptor.methods."setInstancedAccelerationStructures:".unsafe = false
 class.MTLAccelerationStructureGeometryDescriptor.methods."setPrimitiveDataBuffer:".unsafe = false
 class.MTLAccelerationStructureTriangleGeometryDescriptor.methods."setVertexBuffer:".unsafe = false
-protocol.MTLRenderCommandEncoder.methods."useResource:usage:".unsafe = false
-protocol.MTLRenderCommandEncoder.methods."useResource:usage:stages:".unsafe = false
-protocol.MTLComputeCommandEncoder.methods."useResource:usage:".unsafe = false
-protocol.MTLBlitCommandEncoder.methods."synchronizeResource:".unsafe = false
 protocol.MTLBlitCommandEncoder.methods."generateMipmapsForTexture:".unsafe = false
 protocol.MTLBlitCommandEncoder.methods."optimizeContentsForGPUAccess:".unsafe = false
 protocol.MTLAccelerationStructureCommandEncoder.methods."copyAndCompactAccelerationStructure:toAccelerationStructure:".unsafe = false
+
+# SAFETY: Synchronizing a resource is safe, as it:
+# - Doesn't need to the resource to be bound.
+# - Doesn't access the contents of the resource.
+# - Doesn't need the resource to be retained by the user, since it already is
+#   guaranteed to be since they've used the resource elsewhere.
+#   TODO(breaking): Is that claim true?
+protocol.MTLBlitCommandEncoder.methods."synchronizeResource:".unsafe = false
+
+# SAFETY: We consider resource's bound-ness as safe, the safety is moved to
+# command encoding methods such as `setTexture:atIndex:` instead.
+#
+# Calling this with a resource is safe, since it:
+# - Doesn't access the contents of the resource.
+# - Doesn't need the resource to be retained by the user, since it already is
+#   guaranteed to be since they've used the resource elsewhere.
+#
+# TODO(breaking): Are those claims true?
+protocol.MTLAccelerationStructureCommandEncoder.methods."useResource:usage:".unsafe = false
+protocol.MTLComputeCommandEncoder.methods."useResource:usage:".unsafe = false
+protocol.MTLRenderCommandEncoder.methods."useResource:usage:".unsafe = false
+protocol.MTLRenderCommandEncoder.methods."useResource:usage:stages:".unsafe = false
 
 # SAFETY: Resource options are safe to specify:
 # - Hazard tracking and storage modes change the required synchronization, but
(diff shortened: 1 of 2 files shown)


framework-crates/objc2-metal/translation-config.toml
```diff
@@ -33,864 +33,138 @@ enum.MTLTensorDataType.use-value = true
...
-class.MTLVertexAttributeDescriptor.methods.new.unsafe = false
-class.MTLVertexAttributeDescriptor.methods.format.unsafe = false
-class.MTLVertexAttributeDescriptor.methods."setFormat:".unsafe = false
-class.MTLVertexAttributeDescriptor.methods.offset.unsafe = false
-# setOffset: must be a multiple of 4
-class.MTLVertexAttributeDescriptor.methods.bufferIndex.unsafe = false
-# TODO: Verify that index out-of-bounds is sound.
-# setBufferIndex:
+# TODO.
+class.MTLHeapDescriptor.methods."setType:".unsafe = true
```
MarijnS95, line 114: This one was safe, now it is not; what was your plan for it? It might be a breaking change, at least it'll break https://github.com/Traverse-Research/gpu-allocator if released as a patch.

framework-crates/objc2-metal/translation-config.toml
```diff
@@ -33,864 +33,138 @@ enum.MTLTensorDataType.use-value = true
...
-protocol.MTLDevice.methods.supportsPullModelInterpolation.unsafe = false
-protocol.MTLDevice.methods.supportsShaderBarycentricCoordinates.unsafe = false
-protocol.MTLDevice.methods.currentAllocatedSize.unsafe = false
-protocol.MTLDevice.methods.newCommandQueue.unsafe = false
+# SAFETY: These are not bounds-affecting.
+class.MTLAccelerationStructureTriangleGeometryDescriptor.methods."setIndexBuffer:".unsafe = false
+class.MTLAccelerationStructureTriangleGeometryDescriptor.methods."setTriangleCount:".unsafe = false
+class.MTLAccelerationStructureBoundingBoxGeometryDescriptor.methods."setBoundingBoxCount:".unsafe = false
+class.MTLInstanceAccelerationStructureDescriptor.methods."setInstanceCount:".unsafe = false
+class.MTLIntersectionFunctionTableDescriptor.methods."setFunctionCount:".unsafe = false
```
MarijnS95, line 68: Some affect the bounds of GPU buffer reads, right?

madsmtm:
I've looked through [the documentation on Metal resources](https://developer.apple.com/documentation/metal/resource-fundamentals?language=objc), and I think the correct choice is to mark (basically) all resource (`MTLBuffer`/`MTLTexture`/``) accesses as `unsafe`, since the user has to synchronize accesses to these?

I have implemented this in the latest commit, please let me know what you think?

framework-crates/objc2-metal/translation-config.toml
```diff
@@ -33,864 +33,213 @@ enum.MTLTensorDataType.use-value = true
...
+# protocol.MTLResidencySet.methods.requestResidency.unsafe = false
+# protocol.MTLResidencySet.methods.endResidency.unsafe = false
+
+# TODO(breaking): Mark this as unsafe?
+class.MTLHeapDescriptor.methods."setType:".unsafe = false
+
+# These affect lifetime safety, and can cause use-after-free if used incorrectly.
+class.MTLCommandBufferDescriptor.methods."setRetainedReferences:".unsafe = true
+protocol.MTLCommandQueue.methods.commandBufferWithUnretainedReferences.unsafe = true
+protocol.MTLIOCommandQueue.methods.commandBufferWithUnretainedReferences.unsafe = true
```
madsmtm, line 186: @MarijnS95 pointed out in https://github.com/madsmtm/objc2/pull/773#issuecomment-3352377509:

> * **Lifetime safety**: This [used to be tracked by Metal if not opted out](https://developer.apple.com/documentation/metal/mtlcommandqueue/makecommandbufferwithunretainedreferences()?language=objc) but that already wasn't possible when manually uploading `gpuReourceID()` values, and [is removed now in Metal 4](https://developer.apple.com/documentation/metal/understanding-the-metal-4-core-api?language=objc#Command-buffers).

Which is just.. Ugh! Not yet sure how we'll resolve that. Maybe all command buffer and command encoder methods need to be unsafe? Maybe it'll be fine since most of them already are?

MarijnS95: Yeah, all dispatches on the GPU (shader dispatches, draws, copy commands, ...) should likely be unsafe because of resource lifetimes :shrug:

madsmtm: Re-reading the docs for [`commandBufferWithUnretainedReferences`](https://developer.apple.com/documentation/metal/mtlcommandqueue/makecommandbufferwithunretainedreferences()?language=objc), it says:
> Returns a command buffer from the command queue that doesn’t maintain strong references to resources.

So it sounds like the things that aren't retained only applies to `MTLResource` and subprotocols, and that it'd still keep a reference to any `MTLFence`, `MTLCommandBufferHandler`, `MTLSamplerState` etc. that you pass to the buffer / its encoders?

As such, I think we can again get around this by marking `MTLResource` as unsafe (and probably add another safety requirement to the docs)?

madsmtm:
I think this PR is ready now. Overall, I believe that marking `MTLResource` and `MTLFunction` as `unsafe` was the correct choice, and that this mostly resolves the safety concerns outlined in https://github.com/madsmtm/objc2/pull/773#issuecomment-3218100934.

I've gotten a bit lost between all our threads on this, but I think I've answered all your concerns @MarijnS95? WDYT about this PR in it's current state (if you look past all the bounds-safety stuff)?

(I strongly suspect there's still more to be done here, but I'm honestly reaching my limit of the amount of unsafe code I can review, I find myself glazing over things by now - so while this PR is probably unsound in a few places, I think it's enough of an improvement that it'd be worth it to merge it now anyhow).

madsmtm:
I know you didn't get the time to look through it Marijn, but I'd like to get the v0.3.2 release out, so I'll merge this now. Feel free to still comment here afterwards once you get the time.

PR #805, opened 2025-12-05
objc2: Minor cleanup

andyleiserson:


crates/objc2-encode/src/lib.rs
```diff
@@ -55,8 +55,7 @@ mod encoding_box;
 mod helper;
 mod parse;
 
-// Will be used at some point when generic constants are available
```
madsmtm: Nit: I'd like to keep this comment, as that is still what the purpose of that module is.

madsmtm (approved):
Thanks! And don't worry too much about the Git stuff, as long as `git blame` mostly works I'm not that picky about it.

PR #808, opened 2025-12-22
Use documentation from Apple's documentation bundle

### LINKED ISSUE #309: Autogenerate documentation in `icrate`

madsmtm:
Use the `apple-doc` crate I introduced in 0dd51d5060662288f5c7d900fd1d111c5d2a735f to parse the data that Xcode's documentation viewer uses (this is the same data as on https://developer.apple.com/documentation), and use that in `header-translator` to enhance the documentation we emit.

Fixes https://github.com/madsmtm/objc2/issues/309.

TODO:
- [ ] Finish the implementation, methods don't get this documentation yet, and links are currently to Apple's website instead of to the Rust item.
- [ ] Consider if there's some way we can include Apple's examples / tutorials, to truly make it unnecessary to want to go to Apple's website?
- [ ] Figure out the legal implications of this? We're already redistributing basically the entire Xcode SDK, this would just be taking and redistributing more of Xcode, but maybe there's a difference? Especially given that you _cannot_ interface with Apple's platform without the SDK, and there is not other way to write them, whereas the documentation is maybe more legally protected?
  - I had a discussion with the Rust Foundation about this in September, but they're taking a long time to get back to me.
- [ ] Optimize this, `cargo run --bin header-translator --release -- all` now takes quite a bit longer.

---

As an example of how this would look, compare the documentation for `NSVisualEffectView` [on Apple's website](https://developer.apple.com/documentation/appkit/nsvisualeffectview?language=objc) with what it would be in `objc2-app-kit` after this PR:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -771,6 +771,7 @@ dependencies = [
 name = "header-translator"
 version = "0.1.0"
 dependencies = [
+ "apple-doc",
  "apple-sdk",
  "clang",
  "clang-sys",
--- a/crates/apple-doc/src/fs_json.rs
+++ b/crates/apple-doc/src/fs_json.rs
@@ -206,10 +206,24 @@ pub enum Reference {
 #[serde(deny_unknown_fields)]
 pub struct ReferenceVariant {
     pub url: String,
-    pub traits: Vec<String>,
+    pub traits: Vec<ReferenceVariantTrait>,
 
     #[serde(rename = "svgID")]
     pub svg_id: Option<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Clone)]
+pub enum ReferenceVariantTrait {
+    #[serde(rename = "1x")]
+    X1,
+    #[serde(rename = "2x")]
+    X2,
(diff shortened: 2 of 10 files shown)


madsmtm:
As of Xcode 27, this is starting to become a _bit_ less relevant, Apple seems to have decided to put more stuff into headers (having machine-readable docs is _also_ nice in Swift it turns out, golly, who knew? /snark). At least for Foundation, there's actually extremely few new Foundation APIs, it's basically all just docs additions, but time will tell if that pattern continues to hold for their other frameworks.

See: https://github.com/madsmtm/objc2/pull/860

PR #829, opened 2026-03-17
Fix(CoreWlan): Make [CWConfiguration networkProfiles] return-nullable

### LINKED ISSUE #828: Panic when accessing CWConfiguration::networkProfiles
I recently started using CoreWifi in my program, but am getting panics when deploying it at large. The panic looks like this:

```
Got a panic: panicked at /builds/vendor/deps/objc2-core-wlan-0.3.2/src/generated/CWConfiguration.rs:52:5:
unexpected NULL returned from -[CWConfiguration networkProfiles]
```

Looking into it, we can indeed see that [`CWConfiguration::networkProfiles`](https://github.com/roblabla/MacOSX-SDKs/blob/master/MacOSX.sdk/System/Library/Frameworks/CoreWLAN.framework/Versions/A/Headers/CWConfiguration.h#L54) is defined in the objc2 header like this:

```objc
NS_ASSUME_NONNULL_BEGIN
@interface CWConfiguration : NSObject <NSCopying, NSMutableCopying, NSSecureCoding>
{
// [...]
}
// [...]
@property(readonly, copy) NSOrderedSet<CWNetworkProfile *> *networkProfiles NS_AVAILABLE_MAC(10_7);
```

As a result, objc2 generates bindings "as if" the function could never return a null pointer (since it doesn't have the `nullable` property).

It's unclear to me in what scenario a configuration can be missing a networkprofile, but I know it can happen at least in some cases on macOS 14.

roblabla:
Fixes #828


madsmtm (approved):
Thanks! Will be released in [v0.4](https://github.com/madsmtm/objc2/milestone/8).

